PR #2, opened 2025-09-26
system structure: refactoring code structure according to system architecture

enkerewpo:


docs/source/uapi/action_system.rst
```diff
@@ -141,25 +141,25 @@ Type Safety
 
 .. code-block:: python
 
-   from uapi import create_runtime_manager, set_runtime
+   from robonix.uapi import get_runtime, set_runtime
    
-   # 创建运行时管理器
-   manager = create_runtime_manager()
+   # 获取运行时实例
```
ken4647, line 146: Should we translate the comments into English?

.gitmodules
```diff
@@ -1,3 +1,15 @@
 [submodule "robonix/driver/realsense_ros2/src/realsense-ros"]
-	path = robonix/driver/realsense_ros2/src/realsense-ros
-	url = https://github.com/IntelRealSense/realsense-ros.git
+    path = robonix/driver/realsense_ros2/src/realsense-ros
```
ken4647, line 2: Actually, I prefer to manage each cap/driver individually rather than placing them directly in the project root directory and implementing them through gitmodules. On one hand, this increases system coupling, and on the other hand, in some cases, certain ROS packages do not provide git links.Of course, considering that it has been like this before, I plan to open an issue to address this problem.

docs/source/uapi/action_system.rst
enkerewpo: ok

.gitmodules
enkerewpo: I think you mean that we need something like "package manager" that automatically download and build the drivers or caps the user requires (not everyone uses mid360, so we may provide something like `apt install xxx`)

ken4647 (approved):

PR #3, opened 2025-09-28
fix some small issues about thread safety and add requirements.txt

KouweiLee:
I went through the code about action and found some small issues. So I made the following fixes:
- Introduced thread-local storage for managing the current entity in the Runtime class, improving thread safety.
- Add one requirements.txt


examples/demo1/simple_demo.py
```diff
@@ -15,9 +15,9 @@
 sys.path.insert(0, str(project_root_parent))
 
 from robonix.uapi.log import logger
-from robonix.uapi import create_runtime_manager, set_runtime
+from robonix.uapi import create_runtime_manager, set_runtime, RuntimeManager, Runtime
```
enkerewpo, line 18: the `RuntimeManager` is deprecated now

PR #9, opened 2025-10-29
add find chair example and llm skill

lhw2002426:


robonix/capability/nav2_webots_tiago/plugins/ros2_navigation/lib.py
```diff
@@ -74,6 +74,6 @@ def set_goal(x, y, yaw) -> str:
     return func_status
     
 if __name__ == "__main__":
-    rclpy.init()
+    # rclpy.init()
```
ken4647: Can this line be removed directly?

PR #17, opened 2026-01-08
Fix the issue where run.sh deletes the container upon exit after creating it from an image each time.

LittleRookie1115:
## Description
This PR modifies `run.sh` to support iterative development in a local container by making two key changes:

1. **Removes the `--rm` flag** (line 104) from the `docker run` command, allowing the container to persist after exit for inspection and reuse.
2. **Adds a check for existing running containers** with the same name before creating a new one. This prevents errors and allows developers to repeatedly modify and test within the same local container environment.

These changes are especially helpful for developers who are new to the project, as they enable a stable, long-term development environment within a single container, eliminating the need to rebuild from scratch for every test.


enkerewpo:
The CI will check commit messages format according to https://www.conventionalcommits.org/en/v1.0.0/#summary
(Commits starting with "Merge ..." are ignored)

You can rename your commit with "fix: ..." to pass the PR CI checking

PR #32, opened 2026-05-06
Dev liaison

kaileliu:
Robonix integration with Liaison voice interaction

Built on the Robonix dev branch, this work wires Liaison’s multimodal input path: text is forwarded through Liaison; voice is handled via ASR, TTS, and related services under a single Liaison orchestration layer; TUI tests pass.
Voiceprint (speaker) identification is not included yet and can be added later.
This change adds voice interaction to Robonix. Previously users could only talk to the system via text; now, in the TUI, Ctrl+V runs the full pipeline record → speech recognition → reasoning → speech synthesis automatically.


enkerewpo (approved):
OK

enkerewpo:
I will take it from here and rebase the code to `dev` later since there have been some code structure and api changes, so I raw merged it into the tmp branch `dev-liaison` just now.

You can also commit to `dev-liasion` directly in `syswonder/robonix` later, and there will be another PR for `dev-liaison` to `dev`

PR #38, opened 2026-05-16
Update RTDL plan structure and add parallel operation

KouweiLee:
I made some changes in this pr:
1. To support parallel operator, I changed Plan.msg and added RtdlNode.msg to represent the rtdl plan for executor.
2. I moved the rtdl-related prompt to rtdl_protocol.md, so this md acts both as the prompt to LLM and the doc for RTDL protocol.


1mujue (approved):
Got it

PR #47, opened 2026-05-22
feat(voiceprint): add ECAPA-TDNN voiceprint service

kaileliu:
Adds system/voiceprint/ Python service implementing the robonix/system/speech/voiceprint capability:

- EcapaTdnnEngine wrapping SpeechBrain spkrec-ecapa-voxceleb (192-dim embeddings, cosine-similarity matching).
- gRPC service with Identify, plus package-private Enroll / ListEnrolled admin RPCs on the same listener.
- uv-managed venv + grpc_tools codegen via scripts/build.sh, with HuggingFace mirror pre-warming the model cache.
- unittest E2E suite under tests/ exercising enroll/identify/reject against bundled wav fixtures.

The capability TOML and ROS IDL (Identify.srv) are untouched; Enroll/ListEnrolled are package-private and not registered as Atlas capabilities.


system/voiceprint/proto/voiceprint.proto
enkerewpo, line 1: please use capability tomls + `rbnx codegen` to generate the gRPC protos, not directly write one

KouweiLee (approved):

PR #78, opened 2026-06-08
feat(scene): restore stable objects across restarts

HeartLinked:


enkerewpo:
## Review + e2e test report

Tested this PR rebased onto the latest `dev` in the Webots example (full stack: sim + rbnx boot + rviz, RTX 2060), driving scene through `rbnx ask` (pilot → executor → scene MCP).

### What works ✅
- **Perception**: `ConceptGraphsDetector` (rgb+depth) runs continuously; scene graph populates and stays stable across hundreds of ticks (detected the potted plant / cabinet / etc. from the sim world).
- **Persistence mechanism**: `ObjectStore` (milvus-lite, per-`map_id`, composite PK `{map_id}::{object_id}`) initializes and the restore-on-startup path is wired correctly behind `SCENE_OBJECT_MEMORY_ENABLED`.
- **End-to-end query**: `list_objects` over MCP returns well-formed objects with coordinates (`cabinet x=2.63 y=-0.56 z=0.28 yaw=1.09 ...`) through the full pilot→executor→scene chain.

### Main concern — restore is not spatially meaningful yet ⚠️
Persisted objects are stored in the **`map` frame**, keyed by `map_id`. But the `map` frame origin is established from the robot's start pose at each boot, and `mapping` does **not** yet support map save/load + `map_id`. So objects restored from a previous session are expressed in a *different* coordinate frame than the new session's `map` → restored positions do not align.

In other words: **warm-restore of the semantic map is only meaningful once the spatial map is itself persisted and reloaded under the same `map_id`.** Until then, restore must stay off.

Good news: restore is already gated behind `SCENE_OBJECT_MEMORY_ENABLED` (default **off**), so this ships as a dormant capability and won't silently restore stale coordinates. 

**Request before merge:** document this dependency explicitly (README / persistence.py) — that `SCENE_OBJECT_MEMORY_ENABLED` should remain off until `mapping` provides stable, reloadable map frames bound to `map_id`. Right now the docs don't state the precondition.

### Minor
- `docker/requirements.txt` pins `pymilvus!=2.6.10,>=2.5.0` with **no upper bound**. A fresh image build resolved `pymilvus==3.0.0` (the code is written against the 2.x `MilvusClient` API; memsearch elsewhere resolves 2.6.12). Recommend capping `<3` to avoid drift.

### Notes (not blockers, not this PR)
- Two unrelated issues surfaced during the run and are **not** caused by this PR: (1) pilot VLM key expiry; (2) on `dev`, the executor routes loopback MCP (`127.0.0.1`) through `http_proxy`, returning 502 — fixed by a `no_proxy` change on the `executor_pilot` branch, absent on `dev`. Worked around here with `no_proxy=127.0.0.1`.

Functionally the scene-side code is sound. The only merge-gating ask is documenting the map-frame precondition for restore.

PR #80, opened 2026-06-09
docs(readme): add contributors section via all-contributors

enkerewpo:
## What

Adds a **Contributors** section to the README, just above `## License`, rendered with the [all-contributors](https://github.com/all-contributors/all-contributors) spec — avatars + real (display) names, 5 per row.

Also adds `.all-contributorsrc` at the repo root so the list is maintainable going forward.

## Why all-contributors (vs contrib.rocks / an auto Action)

`contrib.rocks` and the contributor-graph Actions auto-populate but can only show **usernames** — the GitHub contributor API doesn't carry reliable real names. all-contributors keeps a curated list, which is the only way to show real names. Trade-off accepted intentionally.

## English-only docs constraint

The repo's `Docs check (English-only)` CI (`scripts/check-docs.sh`) forbids CJK in in-repo docs, so contributor display names must be romanized. Two names were latinized accordingly (`Zhaobo Zhang`, `Kaile Liu`); the rest were already in Latin script. Verified locally against the CI's own CJK detector before pushing.

## Maintaining the list

The list is curated in `.all-contributorsrc`. To add someone later, either:
- install the all-contributors GitHub App and comment `@all-contributors please add @user for code`, or
- run `npx all-contributors add <user> code && npx all-contributors generate`.

Either path fetches that person's name + avatar automatically and updates the table between the `ALL-CONTRIBUTORS-LIST` markers. Note: if a new contributor's profile name is in CJK, romanize it by hand or the English-only check will fail.

## Notes

- Seeded with the 10 current contributors (bot account excluded).
- `lhw2002426` has no public profile name, so it falls back to the username.
- The `## Skills` section is intentionally left untouched — a separate PR will cover that.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


enkerewpo:
@all-contributors please add @qingfeng123 for Robonix-Skill-Toolkit

enkerewpo:
@all-contributors please add @QingFeng34048 for Robonix-Skill-Toolkit

enkerewpo:
@all-contributors please add @QingFeng34048 for tool, data

PR #89, opened 2026-06-18
feat(scene): hardware-aware perception tiers with contract-sourced camera intrinsics

HeartLinked:
Depth back-projection scales every 3D point by fx/fy and offsets it by cx/cy. Feeding ConceptGraphs incorrect intrinsics silently places objects at the wrong map position.
Scene previously defaulted every camera to the Webots Tiago K intrinsics: That default is correct for simulation only. Any other camera produced misplaced objects without crashing or warning unless an operator manually set SCENE_CAMERA_INTRINSICS.

Detector selection was previously implemented as an inline if/elif on hub.has(...), which silently fell through to detector disabled when no camera was wired.This change extracts detector selection into a pure, testable probe that names the tier Scene runs and logs the plan once at startup.

Adds plan_perception(hub), returning a frozen PerceptionPlan that selects one of three tiers based on camera availability:
metric: RGB-D camera → ConceptGraphs
visual: RGB camera → VLM
geometric: no camera → detector disabled; occupancy_grid and goal_near remain available

### DIFF
--- a/capabilities/primitive/camera/intrinsics.v1.toml
+++ b/capabilities/primitive/camera/intrinsics.v1.toml
@@ -1,0 +1,32 @@
+# Static camera intrinsics (pinhole K + image size). The camera
+# primitive declares this so downstream consumers (scene's depth
+# back-projection, any 3D reconstruction) get the *real* per-deployment
+# intrinsics instead of guessing.
+#
+# Why this is a contract and not a hardcoded default:
+#   Depth back-projection scales every 3D point by fx/fy and offsets it
+#   by cx/cy. A consumer that assumes one camera's intrinsics (e.g. the
+#   Webots head camera) and is fed another camera's depth (a RealSense
+#   at a different resolution / FOV) places objects at systematically
+#   wrong map positions — silently, with no crash. Routing K through
+#   atlas means the consumer asks for `primitive/camera/intrinsics` and
+#   gets the geometry of whichever camera is actually wired in.
+#
+# Payload: sensor_msgs/CameraInfo — `k` is the 3x3 row-major intrinsic
+# matrix (`k[0]=fx`, `k[2]=cx`, `k[4]=fy`, `k[5]=cy`); `width`/`height`
+# are the image dimensions the matrix is calibrated for. These describe
+# the same optical frame as `primitive/camera/extrinsics` publishes
+# (i.e. the camera the RGB/depth streams come from).
+#
+# Cadence: TRANSIENT_LOCAL + RELIABLE (latched). Intrinsics are static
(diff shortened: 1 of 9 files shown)


enkerewpo:
## Review

Strong PR — clean design, defensive code, and well-documented. Sourcing intrinsics from a contract instead of a hardcoded default is the right call and directly removes a class of silent "objects placed at the wrong map position" bugs.

Verified the capability probe in isolation: ran `tests/test_capabilities.py` (pure Python, no ROS/atlas) — **5/5 pass** (metric / metric-degraded / visual / geometric / summary). The `plan_perception` extraction into a frozen `PerceptionPlan` is a nice testability win over the old inline `if hub.has(...)`.

### ⚠️ Behavioral change to confirm before merge
The metric path no longer has a hardcoded-K fallback: if no usable `primitive/camera/intrinsics` is wired, `_cam_info()` returns `None` and ConceptGraphs **waits** (`waiting for camera intrinsics`) — it produces **no objects** until intrinsics land. This is the correct trade (no silent mis-grounding), but it raises a hard requirement: **every deployment's camera primitive must now publish the intrinsics contract, or scene goes silent on the metric tier.**

- Webots is covered — this PR's `tiago_camera/driver.py` relays `/head_front_camera/rgb/camera_info` onto the latched contract topic, and the sim does publish `camera_info`.
- **Please confirm the real-robot camera primitive (MID-360 rig / RealSense) actually publishes `primitive/camera/intrinsics`.** If it doesn't, merging this makes the car's scene stop detecting objects (only a `WARN` + `waiting for camera intrinsics` in the logs). That's the one thing I'd nail down before merge.

### Minor: `grounding` label conflates two very different states
`PerceptionPlan.grounding` returns `"degraded"` when *either* intrinsics *or* pose is missing. But those degrade very differently:
- missing **pose** → falls back to tf2, **still produces objects**;
- missing **intrinsics** → detector **waits forever, zero output**.

Folding "works-but-coarse" and "produces-nothing" under one `degraded` label undersells the second. Consider distinguishing them (e.g. `degraded-pose` vs `no-intrinsics`) in the label / startup log so an operator can tell "grounding is rough" from "scene will never emit an object."

### Merge-order note vs #78
Both this PR and #78 (object persistence) edit `system/scene/scene_service/service.py` and `README.md`. Different concerns (perception dispatch vs persistence wiring), but they'll conflict on merge — whichever lands first, the other needs a rebase.

### Verdict
Code is sound and better-tested than most scene changes. Not blocking on style. The only real gate is confirming the real-robot intrinsics publisher exists; with that confirmed (or a documented fallback for cameras that can't publish CameraInfo), this is good to merge.

PR #91, opened 2026-06-20
feat: add soma support

Chenyantt:
Add soma system service support, including:
# Soma — raw body YAML and URDF service

Soma loads one or more deployment directories, caches each deployment's raw
`soma.ymal` and referenced URDF text, then exposes them to other Robonix
components over gRPC.

This simplified version does not render or interpret a self-description. Pilot
calls `robonix/system/soma/get_ymal` to receive the original YMAL string, and
other services call `robonix/system/soma/get_urdf` to receive the corresponding
URDF XML string.

## Config

```yaml
atlas_endpoint: 127.0.0.1:50051
listen: 127.0.0.1:50091
provider_id: soma
robonix_root: ../../
default_robot: test_ci_robot
start_packages: false
rbnx_bin: rbnx
deployments:
  - examples/test_ci
```

Relative `robonix_root` is resolved from the config file directory. Relative
`deployments` entries are then resolved from that Robonix root. For example,
when the config file is `system/soma/config.ymal`, `robonix_root: ../../` and
`deployments: [examples/test_ci]` point at `<repo>/examples/test_ci` no matter
which directory `cargo run -p robonix-soma -- --config <path>` is launched from.
If `robonix_root` is omitted, Soma tries `ROBONIX_SOURCE_PATH`, `ROBONIX_ROOT`,
then searches upward from the current directory for the Robonix repository root.

Each deployment directory must contain:

- `robonix_manifest.yaml`
- `soma.ymal` or compatible names such as `robonix.soma.yaml`
- the URDF file referenced by `urdf.path`

## Run

```bash
cargo run -p robonix-soma -- --config ./soma.local.yaml
```

At startup Soma reads every deployment, optionally runs each local primitive and
skill package with `rbnx start -p <package_dir> --endpoint <atlas>`, prints a
startup report, registers itself in Atlas, and serves gRPC on `--listen`.

### DIFF
--- a/services/memsearch/memsearch_service/service.py
+++ b/services/memsearch/memsearch_service/service.py
@@ -60,7 +60,7 @@ logging.getLogger("absl").setLevel(logging.ERROR)
 
 # Transitional: also emit key lifecycle events through Scribe so the
 # unified `rbnx logs` tool can see them alongside other components.
-scribe_logger.info("service_memory", "memory service starting")
+scribe_logger.info("memory", "memory service starting")
 
 def _log_environment() -> None:
     """Print a compact environment summary at boot. Many of the reported
--- a/system/atlas/src/main.rs
+++ b/system/atlas/src/main.rs
@@ -29,23 +29,29 @@ struct Args {
     #[arg(long, env = "ROBONIX_ATLAS_CAPABILITIES", value_delimiter = ',')]
     capabilities: Vec<String>,
 
-    /// Log filter (env_logger syntax: `info`, `robonix_atlas=debug`, …).
-    /// Default: `robonix_atlas=info`. Also reads `RUST_LOG` if set.
+    /// Log level for this component (`debug`/`info`/`warn`/`error`). Sets the
+    /// scribe log-file floor; falls back to `SCRIBE_FILE_LEVEL` / `info`.
+    /// Normally arrives inside `--config-json`, not as a standalone flag.
     #[arg(long)]
     log: Option<String>,
+
+    /// The component's `system.atlas` manifest block, serialized to JSON by
+    /// rbnx and passed as one arg (`--config-json '{…}'`). Parsed by the binary
+    /// itself — see `robonix_scribe::init_from_config`, which reads the `log`
+    /// key from it so the manifest's per-component level reaches the log.
+    #[arg(long)]
+    config_json: Option<String>,
 }
 
 #[tokio::main]
 async fn main() {
     let args = Args::parse();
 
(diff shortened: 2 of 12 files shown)


system/soma/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+use anyhow::{Context, Result};
+use clap::Parser;
+use robonix_atlas::client::{self as atlas_client, AtlasClient};
+use robonix_atlas::pb as atlas_pb;
+use robonix_soma::config::{Args, SomaConfig};
+use robonix_soma::deployment::DeploymentStore;
+use robonix_soma::launcher::PackageLauncher;
+use robonix_soma::pb::contracts::{
+    robonix_system_soma_get_urdf_server::RobonixSystemSomaGetUrdfServer,
+    robonix_system_soma_get_ymal_server::RobonixSystemSomaGetYmalServer,
```
enkerewpo: should be yaml_server?

Chenyantt: fix it

lhw2002426:
Issue 1 — system_cli_args for soma emits --start-packages false, which clashes with soma's flag-style CLI
In tools/rbnx/src/cmd/deploy.rs, scalar("start_packages") translates the manifest's start_packages: false into the CLI arg --start-packages false. But soma defines --start-packages as a clap flag (no value — see system/soma/src/config.rs:59). The false is parsed as a stray positional arg and soma exits with unexpected argument 'false', so rbnx boot never gets soma to register.

Two ways to fix:

(a) Change soma to --start-packages <bool> value-style (action = ArgAction::Set, num_args = 1).
(b) Change rbnx to only push --start-packages when the value is true, and omit it otherwise.

lhw2002426:
Issue 2 — normalize_path turns a lone . into an empty PathBuf
In system/soma/src/config.rs, normalize_path drops every Component::CurDir. When a user writes the natural single-deploy config (robonix_root: ., deployments: [.]), normalization yields an empty PathBuf and soma fails at startup with no Soma YAML file found in ''.

Fix: after the component loop, push . if out is empty; or anchor relative inputs to std::env::current_dir() before the loop.

Chenyantt:
done

enkerewpo:
please fix ci

Chenyantt:
done

examples/webots/soma.yaml
enkerewpo, line 1: all `/tiago_webots/*` in `exports` do not match current capability contract naming convention, please change to `robonix/primitive/camera/xxx`, `robonix/service/xxx`, `robonix/skill/explore/xxx`

PR #92, opened 2026-06-21
feat: RTDL forest execution, mid-task steering, and live task graph in rbnx chat

enkerewpo:
## Summary

Brings the `executor_pilot` line into `dev`: the **RTDL (Robot Task Description Language) execution model** plus the Pilot planning/steering layer and a live `rbnx chat` visualization. Two layers:

### Executor — RTDL execution runtime (streaming)
- Interprets a Pilot `Plan` (an RTDL node arena of `sequence` / `parallel` / `do`) and streams `RtdlEvent`s back (`plan_started` / `node_state` / `plan_complete`).
- Async capability polling (sibling `status` contract) for long-running caps; `cancel_plan` + `cancel_all_plans` builtins; per-node `op_id` + `description`.

### Pilot — planning, steering, forest, task state
- The VLM emits a single JSON envelope per round: `{content, rtdl_description, rtdl, task_update}` (no OpenAI tool-calls). `task_update` is the LLM-owned overall goal `{goal, success_criterion, status}`; `status: "done"` is the one authoritative completion signal.
- `run_turn` is a non-blocking **forest supervisor loop**: each dispatched RTDL tree runs in its own task, so trees dispatched across rounds run concurrently (the forest). A `select!` loop reacts to tree events, mid-task steering, and cancellation.
- **Steering**: a Task submitted while a turn is live is routed into that turn's queue (not a second turn). The LLM sees the in-flight forest in its prompt and decides for itself whether to `builtin_cancel_plan`.
- Plan IDs are pilot-assigned, monotonic per session (the LLM never picks them).
- **Invalid-RTDL recovery**: one corrective retry, then a graceful turn end instead of crashing the stream.
- **Latency**: full protocol only on round 0 (a precise reminder thereafter) + Claude-Code-style rolling history compaction once the window grows large.

### rbnx chat
- Mid-task input: type + Enter while a task runs to send a steer.
- Right-side live **Task / Forest** panel: current goal + status, and every in-flight RTDL tree as an ASCII tree with per-node state markers, updated in real time.

## Contract changes
- `PilotEvent` gains `node_state` / `task_state`; new `PilotNodeState` / `TaskStateEvent` messages. These are Pilot-internal output (relayed verbatim by Liaison); the Executor contracts are unaffected.

## Merged from dev
[...]

PR #93, opened 2026-06-21
feat(scribe): add scribe logs

ohhhHwH:
a lightweight facade that replaces scattered logging with one consistent format, routing, and viewing surface across all Robonix processes.


ohhhHwH:
2bbf0addb0e8b0301bb07d26aaf7fb9c
image
image
Here's a screenshot of the Scribe logs

system/executor/src/main.rs
```diff
@@ -32,16 +31,15 @@ use std::time::Duration;
...
-        .or_else(|| std::env::var("RUST_LOG").ok())
-        .unwrap_or_else(|| "robonix_executor=info".to_string());
-    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or(log_filter)).init();
+        .or_else(|| std::env::var("RUST_LOG").ok());
+    robonix_scribe::info("executor", "robonix-executor starting");
 
     let cfg = ExecutorConfig::resolve(parsed)?;
 
-    info!("connecting to atlas at {}", cfg.atlas_endpoint);
+    log::info!("connecting to atlas at {}", cfg.atlas_endpoint);
```
enkerewpo:
```
robonix_scribe::info("executor", "robonix-executor starting");
log::info!
```

why two types of log info function? we should remove the previous log crate and use scribe crate as the `info!` `warn!` etc. macros

system/liaison/examples/mock_audio.rs
```diff
@@ -185,7 +185,7 @@ impl RobonixPrimitiveAudioSpeaker for MockSpeaker {
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info")).init();
+    robonix_scribe::info("mock_audio", "mock audio starting");
```
enkerewpo: robonix_scribe::info is too long, please make them as macros just like before

system/liaison/src/main.rs
```diff
@@ -389,12 +389,8 @@ struct Args {
...
 async fn main() -> Result<()> {
     let args = Args::parse();
-    let log_filter = args
-        .log
-        .clone()
-        .or_else(|| std::env::var("RUST_LOG").ok())
-        .unwrap_or_else(|| "robonix_liaison=info".to_string());
-    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or(log_filter)).init();
+    let _ = args.log.clone().or_else(|| std::env::var("RUST_LOG").ok());
+    robonix_scribe::info("liaison", "robonix-liaison starting");
```
enkerewpo: currently, the `tag` field should be the capability provider id if the process is a capability provider. However, let the programmer write "liaison" each time using the info function is definitely not good. We should provide some `init` function that can set all logs from this process with prefilled tags as its capability provider id string or whatever, thus we won't need to write the tag each time

ohhhHwH: Thanks for the tip, it's been changed now, it's been changed，

ohhhHwH:
now, `use robonix_scribe::{info, warn};` imports the macros info! and warn! into the current scope；`robonix_scribe::init("executor");` Used to initialize the log output for the current file；`info!("xx");warn!("xxx");`Used to record logs

enkerewpo (approved):
LGTM with some minor comments

enkerewpo:
## Review

Read the core (`scribe/src/{lib,sink,format}.rs`, `scribe_logger.py`, `rbnx logs`) and the call-site rewrites. Clean single-entry design (`log(level, tag, msg)`), per-sink level filtering with sensible defaults (console=Warn, file=Info), good test coverage (serde roundtrip + legacy-int compat, concurrent file writes), and the Python facade mirrors the Rust format. CI is green (fmt+clippy+build+tests). Nice work.

One thing I'd want fixed before this becomes the logging path for **every** process, plus a few minors.

### ⚠️ Should-fix: the Rust facade panics on I/O failure
Scribe gets threaded into atlas / executor / pilot / liaison / rbnx. A logging facade must never be able to crash its host — but the Rust side has three panic sites on the I/O path:

1. `FILE_SINK` = `LazyLock<FileSink>` … `.expect("failed to create scribe log dir")` (`lib.rs`) — first `log()` call panics the process if `$SCRIBE_LOG_DIR` is unwritable (read-only rootfs, full disk, bad mount).
2. `sink.rs` `write()`: `OpenOptions::…open(&path).expect("failed to open log file")` — panics if a tag's file can't be opened.
3. `init()` panics if called twice.

Compare the **Python** side, which is correct: both sinks are wrapped in `try/except Exception: pass` ("disk full silently ignored"). The Rust side should match — on sink failure, degrade to console-only (or drop), never `panic!`. As written, a read-only `SCRIBE_LOG_DIR` on the robot would take down the whole control plane on its first log line. (For `init()`, prefer set-or-ignore over panic-on-second-call.)

### Minor
- **Tag isn't sanitized before use as a filename** (`sink.rs` `tag_path`: `{dir}/{tag}.log`; same in `_get_writer`). A tag containing `/` or `..` yields an invalid path / traversal and — combined with the `.expect` above — a panic. Sanitize (e.g. replace non-`[A-Za-z0-9._-]`) before forming the path.
- **No log rotation.** `{tag}.log` is append-only and unbounded; a long-running robot session will grow these without limit. Fine for now, but worth a follow-up (size/age cap or external logrotate note).
- **Local-time timestamps, no TZ offset.** `ts_fmt`/`_ts_fmt` serialize local time with no offset, and `ts_parse` re-interprets via local `mktime` — so a log file written in one TZ and read in another shifts (the `log_record_roundtrip` test even allows a 1-day slop). Consider UTC-on-disk, or store the offset, for cross-machine correlation.

### Coordination: large overlap with #92 (`executor_pilot`)
The two PRs touch **13 of the same files**, including `pilot/{planner,discovery,service,memory}.rs`, `executor/{service,main}.rs`, `atlas/{client,service}.rs`, `rbnx/{chat,main}.rs`, `liaison/voice.rs` — both rewrite logging / surrounding lines in each. Whichever merges first, the other needs a non-trivial rebase. Worth deciding the order explicitly with @kouweilee / the executor_pilot work so it's a clean resolution rather than a scramble.

### Verdict
Design and tests are solid and CI is green. The panic-safety fix is the one real gate — once a facade is in every process's hot path, "logging can crash the process" is the failure mode you least want. The rest are minor / follow-up.

ohhhHwH:
Thank you for the review. I'll fix the issues as soon as possible

ohhhHwH:
Hi, I've fixed the two P0 bugs (I/O path panic safety and tag sanitization). The P1 features (log rotation and UTC timestamps) will be submitted in a follow-up PR.

enkerewpo:
@all-contributors please add @ohhhHwH for code

PR #107, opened 2026-06-27
feat: add soma support

Chenyantt:
Add soma system service support, including:
# Soma — raw body YAML and URDF service

Soma loads one or more deployment directories, caches each deployment's raw
`soma.ymal` and referenced URDF text, then exposes them to other Robonix
components over gRPC.

This simplified version does not render or interpret a self-description. Pilot
calls `robonix/system/soma/get_ymal` to receive the original YMAL string, and
other services call `robonix/system/soma/get_urdf` to receive the corresponding
URDF XML string.

## Config

```yaml
atlas_endpoint: 127.0.0.1:50051
listen: 127.0.0.1:50091
provider_id: soma
robonix_root: ../../
default_robot: test_ci_robot
start_packages: false
rbnx_bin: rbnx
deployments:
  - examples/test_ci
```

Relative `robonix_root` is resolved from the config file directory. Relative
`deployments` entries are then resolved from that Robonix root. For example,
when the config file is `system/soma/config.ymal`, `robonix_root: ../../` and
`deployments: [examples/test_ci]` point at `<repo>/examples/test_ci` no matter
which directory `cargo run -p robonix-soma -- --config <path>` is launched from.
If `robonix_root` is omitted, Soma tries `ROBONIX_SOURCE_PATH`, `ROBONIX_ROOT`,
then searches upward from the current directory for the Robonix repository root.

Each deployment directory must contain:

- `robonix_manifest.yaml`
- `soma.ymal` or compatible names such as `robonix.soma.yaml`
- the URDF file referenced by `urdf.path`

## Run

```bash
cargo run -p robonix-soma -- --config ./soma.local.yaml
```

At startup Soma reads every deployment, optionally runs each local primitive and
skill package with `rbnx start -p <package_dir> --endpoint <atlas>`, prints a
startup report, registers itself in Atlas, and serves gRPC on `--listen`.


examples/webots/soma.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - { path: /tiago_webots/navigation/navigate, description: "Navigate to a 2D goal in the mapped  scene." }
+    - { path: /tiago_webots/navigation/status, description: "Report the current navigation goal status." }
+    - { path: /tiago_webots/navigation/cancel, description: "Cancel the active navigation goal." }
+    - { path: /tiago_webots/explore, description: "Run autonomous frontier exploration when the explore skill package is available." }
+  components:
+    - id: base
+      type: mobile_base
+      urdf_link: base_link
+      exports:
+        - { path: /tiago_webots/base/move, description: "Command chassis motion through Webots ROS2 velocity control." }
```
enkerewpo: please use robonix's capability contract convention for `exports`:
`robonix/primitive/chassis/xxx`
`robonix/primitive/camera/xxx`
`robonix/skill/xxx/xxx`

enkerewpo: and why there's no `provider_id` field? each capability belongs to a specific capability provider, identified as `provider_id`:

```yaml
exports:
  - provider_id: tiago_webots_chassis
    capabilities:
      - { path: robonix/primitive/chassis/move, description: "xxx" }
      - { ... }
# for the root node, we have multiple skill providers:
  - provider_id: skill_table_top_1
    capabilities: ...
  - provider_id: skill_exlpore_room
    capabilities: ...
```

capabilities/lib/soma/srv/GetUrdf.srv
```diff
@@ -0,0 +1,7 @@
+# Return the URDF text referenced by one loaded robot's Soma YAML.
+string robot_id  # empty = default robot
+---
+string robot_id
+string package_path
```
enkerewpo: what is this "package path"?

enkerewpo:
@Chenyantt @lhw2002426 please test this PR on webots and also on ranger mini v3 robot

enkerewpo:
## Review — factor process orchestration out of soma, reuse rbnx's `ProcessManager`

**Architecture is correct.** Loading the device-tree `soma.yaml` and launching the primitives/skills bound to it *is* soma's job — no scope problem.

**The real issue: soma re-implements process orchestration that rbnx already owns, but weaker.** `launcher.rs` + `supervisor.rs` are a second, simpler copy of logic that already lives, production-tested, in **`robonix-cli`'s `process` module** — and it's *already exported* (`tools/rbnx/src/lib.rs:11` → `pub mod process`). `ProcessManager` already has everything soma's copy is missing: `kill_process_tree` (SIGTERM → grace → SIGKILL over the whole process group via `killpg`, with `pgrep` child enumeration + fallbacks — real teardown); async `start_process` with stdout/stderr piped to per-package log files + `wait`; process-tree / liveness queries; and logging already routed through `robonix_scribe`.

### Recommended fix (rbnx needs NO change — the module is already pub)
1. soma `Cargo.toml`: add `robonix-cli = { path = "../../tools/rbnx" }`.
2. Delete `system/soma/src/launcher.rs` and `system/soma/src/supervisor.rs`.
3. Use `robonix_cli::process::ProcessManager` to spawn (`start_process`) and tear down (`kill_process_tree`) the device-tree packages.
4. `main.rs`: on a shutdown signal, call the manager's teardown so children don't orphan.

This collapses the weak copy into the maintained one: process bugs get fixed once, and soma inherits process-group teardown + scribe logging for free.

### Concrete bugs in the current soma orchestration (most vanish once you reuse ProcessManager)
1. **[critical] `spawn()` success treated as healthy** (`supervisor.rs:41-51`, `launcher.rs:29-40`): a child that execs then exits a millisecond later (bad manifest / missing dep / port in use) is reported `[OK] pid=…` and soma serves as if up. Fix: grace-window `try_wait` before classifying OK.
2. **[critical] children never reaped / killed** (no `kill_on_drop`, `main.rs` has no shutdown): a soma restart orphans the previous generation to init; a child that dies at runtime is never observed → zombie. `ProcessManager`'s group-kill teardown solves this.
3. **[critical] startup failure never affects health/exit** (`main.rs`, `report.rs`): an all-`SpawnFailed`/`MissingManifest` bring-up still prints `ready`, registers, `SetLifecycleState(ACTIVE)`, exits 0 — a green light over a broken boot. Fix: after the report, if `has_failures()` then bail or reflect a degraded lifecycle state.
4. **[high] no scribe** (`Cargo.toml` has no `robonix-scribe`; raw `eprintln!`; `Args::log` parsed but never read so `--log` is dead). Reusing process.rs gives scribe there; still add `robonix-scribe` + `init` for soma's own lines and wire `--log`.
5. **[medium] `rbnx_bin` precedence inverted** (`config.rs:145` `file_cfg.rbnx_bin.unwrap_or(args.rbnx_bin)` — the file wins over CLI, opposite of every other field; the clap `default_value` also masks an explicit `--rbnx-bin`). Fix: `args.rbnx_bin.or(file_cfg.rbnx_bin).unwrap_or_else(|| "rbnx".into())` and drop the clap default. (Also: `start_packages` defaults to `true` in code vs `false` in the README/config example — pick one; `false` is the safe default.)
6. **[medium] heartbeat / `SetLifecycleState(ACTIVE)` failures swallowed** (`main.rs:80-103`): logged once and ignored; soma keeps reporting "ready" while atlas expires / never marks it (split-brain). Fix: escalate after N failures; make "ready" conditional on ACTIVE actually set.
7. **[medium] error classification by substring** (`service.rs:22-31` matches `msg.contains("unknown Soma robot_id")`): rewording `store.rs` messages silently downgrades NotFound → Internal. Fix: return a typed `StoreError` enum and map variants to gRPC status.

### Solid as-is
`SomaStore` concurrency is clean (immutable `Arc`, read-only handlers, no locks); file reads use `with_context`/`bail!` (no silent skips); consistent `anyhow`, no `unwrap`/`panic` outside tests.

enkerewpo:
The config side needs rework before this can go in.

**The current config is wrong in shape.** soma reads a standalone config file (`robonix_root`, `deployments`, `rbnx_bin`, `default_robot`, `start_packages`). None of that should exist. No other system component has its own config file — atlas/executor/pilot/liaison are all configured from the manifest's `system.<name>` block, which `rbnx boot` serializes to JSON and hands the binary as `--config-json` (plus the typed `--listen`/`--atlas`/`--log`); see `tools/rbnx/src/cmd/deploy.rs:1253-1263` and `system/pilot/src/config.rs:55-96`. And the fields themselves don't hold up: `robonix_root`/`deployments` assume the deploy lives inside the source tree, but `rbnx boot` already runs *in* the deploy dir (which can be anywhere on disk) and serves exactly that one deploy — there's no root to resolve against and no list of deploys. `rbnx_bin` shouldn't exist (don't shell out to a configured rbnx path). `default_robot` is meaningless for a single-deploy process, and `start_packages` is a toggle for the one thing soma exists to do — it also contradicts itself (code defaults `true`, README `false`).

**Fix:** configure soma from the manifest like everyone else. Parse `--config-json`, drop the file loader in `config.rs`, and the whole `system.soma` block is just:

```yaml
  soma:
    listen: 127.0.0.1:50091
    log: info
    robot_yaml: ./robot.yaml   # soma spec (the components tree), relative to the deploy dir
```

atlas endpoint, provider_id, etc. come in through the standard flags/JSON.

**Most important — don't reinvent the launch flow.** rbnx already has the exact boot sequence we want: spawn the package → wait for it to register its `*/driver` capability on atlas → `Driver(CMD_INIT, config_json)` → for primitives also `Driver(CMD_ACTIVATE)`, while skills stop at INACTIVE so the executor lazy-activates them on the first MCP call. soma must keep that behavior unchanged. The *only* difference is the trigger: instead of rbnx walking the manifest's `primitive:`/`skill:` lists, soma walks its own `robot_yaml` (the components tree) and, for each node, looks the launch info up in the robonix manifest by `provider_id` (which is the manifest entry's `name:`) to get its `path:` and `config:`. So: enumerate from the soma spec, pull the how-to-launch from the manifest, then run the *same* rbnx flow — same order, same INIT/ACTIVATE rules. Please literally reuse rbnx's code here — the process spawn/teardown (`ProcessManager`) and the generic Driver call (`call_driver_cmd`, `deploy.rs:1539-1618`) — instead of a parallel reimplementation in `launcher.rs`/`supervisor.rs`.


---

### Appendix A — references

- Developer guide: https://robonix.syswonder.org/developer-guide.html
- Packaging & deployment spec: https://robonix.syswonder.org/integration-guide/packaging-spec.html

### Appendix B — full deploy manifest with soma

`rbnx boot` runs *in* the deploy dir (anywhere on disk). soma is just another `system` component; built-in services (memory/speech) are `service` entries referenced by their robonix-source path.

```yaml
manifestVersion: 1
name: my-robot-deploy
env:

system:
  atlas:    { listen: 0.0.0.0:50051, log: info }
  soma:
    listen: 127.0.0.1:50091
    log: info
    robot_yaml: ./robot.yaml          # soma spec for this deploy (Appendix C)
  executor: { listen: 127.0.0.1:50061, log: info }
  pilot:
    listen: 127.0.0.1:50071
    log: info
    vlm: { upstream: ${VLM_BASE_URL}, api_key: ${VLM_API_KEY}, model: ${VLM_MODEL}, api_format: openai }
  liaison:  { listen: 127.0.0.1:50081, log: info }
  scene:    { log: info }

primitive:
  - name: tiago_chassis               # name == provider_id (the join key)
    path: ./primitives/tiago_chassis
    config: { odom_frame: odom, base_frame: base_link }
  - name: tiago_camera
    path: ./primitives/tiago_camera
    config: { fps: 30 }

service:
  - name: memory                      # built-in service: robonix-source path, NOT ../../
    path: /path/to/robonix/services/memsearch
    config: { backend: sqlite }
  - name: nav2
    url: https://github.com/enkerewpo/nav2_wrapper_rbnx
    branch: main
    config: { params_profile: sim }

skill:
  - name: explore
    url: https://github.com/enkerewpo/explore_rbnx
    branch: main
    config: { explore_mode: frontier }
```

### Appendix C — the soma spec (`robot.yaml`) and the exact launch flow

**Scope: soma brings up primitives and skills only — services are not soma's job.** Boot order is: rbnx starts the `system:` components first (atlas → … → soma); soma, once up, starts/initializes this robot's **primitives and skills**; then rbnx starts every non-system `service:` entry. So soma only ever resolves `provider_id`s against the manifest's `primitive:` and `skill:` lists — never `service:`.

`robot.yaml` carries `exports` at the robot level and per `components` node, each with a `provider_id`. That `provider_id` is the **only** link to the manifest. It should reference primitives (on physical-component nodes) and robot-level skills — not services:

```yaml
# robot.yaml — soma manages primitives + skills (services stay with rbnx)
robot:
  id: tiago
  exports:                            # robot-level skills (not bound to one physical part)
    - provider_id: explore
      capabilities:
        - { path: robonix/skill/explore/room, description: "Frontier exploration." }
  components:                         # physical tree; each node exports its primitive(s)
    - id: base
      type: mobile_base
      exports:
        - provider_id: tiago_chassis
          capabilities:
            - { path: robonix/primitive/chassis/move, description: "Velocity control." }
      components:
        - id: head_camera
          type: rgbd_camera
          exports:
            - provider_id: tiago_camera
              capabilities:
                - { path: robonix/primitive/camera/rgb, description: "RGB stream." }
```

**Launch flow soma must implement (identical mechanics to rbnx today, only the trigger changes):**

1. Read `robot_yaml`. Walk `robot.exports` **plus** every `components[].exports` recursively, collecting each `export.provider_id`. That set is exactly the primitives + skills this robot brings up — in tree order.
2. For each `provider_id`, find the matching entry in the robonix manifest by `name == provider_id`, searching the `primitive:` and `skill:` lists only (services are out of scope). The kind is whichever list it was found in. (If a `provider_id` in `robot.yaml` has no primitive/skill entry → hard error; if a primitive/skill entry is never referenced by `robot.yaml` → warn and skip. Don't silently launch or silently drop.)
3. From that manifest entry take `path`/`url` + `config`, and run the **same** rbnx sequence: `rbnx start`-equivalent spawn → wait for the package to register its `*/driver` capability on atlas → `Driver(CMD_INIT, config_json=<entry.config>)`.
4. Then by kind: **primitive** → also `Driver(CMD_ACTIVATE)` (hot immediately); **skill** → stop at INACTIVE (the executor lazy-activates it on the first MCP call — soma must NOT activate skills).

Worked example from the files above: `robot.yaml` exports `tiago_chassis` (under `components[base]`) → manifest `primitive: tiago_chassis` → spawn `./primitives/tiago_chassis`, INIT with `{odom_frame, base_frame}`, then ACTIVATE. It exports `explore` (robot-level) → manifest `skill: explore` → spawn, INIT with `{explore_mode}`, leave INACTIVE.

Reuse, don't re-derive: spawn/teardown via rbnx's `ProcessManager`, the INIT/ACTIVATE calls via `call_driver_cmd` (`deploy.rs:1539-1618`). The whole of `launcher.rs`/`supervisor.rs` collapses into "enumerate robot.yaml primitive+skill exports → resolve in the manifest → call the shared spawn+driver helpers."

**The rbnx side barely changes.** rbnx keeps exactly the boot logic it has today, with one cut: it no longer iterates the manifest's `primitive:` and `skill:` lists — those are now soma's job. After the `system:` components are up (soma among them, which then brings up the primitives + skills), rbnx just starts the remaining non-system `service:` entries, in manifest order, with the same spawn → register → `Driver(CMD_INIT)` → `CMD_ACTIVATE` sequence it already runs. So: rbnx = system + services; soma = primitives + skills. Nothing else about rbnx's flow moves.

enkerewpo:
I will merge this into `dev-soma` for now and you can continue on that branch, I will add you to this repo later. @lhw2002426 please also help fixing in the `dev-soma` branch.

PR #108, opened 2026-06-27
chore:render Webots on NVIDIA dGPU via PRIME offload

HeartLinked:
On Optimus laptops (Intel iGPU + NVIDIA dGPU in prime-select on-demand), the Wayland compositor and Xwayland run on the iGPU, so the Webots sim container defaulted to rendering on the weak Intel UHD — ~0.25x real-time, shared with rviz2. This adds three GLVND/Vulkan PRIME render-offload env vars to the sim service in compose.gpu.yaml so GL/Vulkan bind the (already-injected) NVIDIA vendor libraries and render on the dGPU instead. The vars are overridable and the file is only merged when nvidia-smi is present, so they're no-ops on single-GPU NVIDIA hosts and never reach pure Intel/AMD setups.

### DIFF
--- a/examples/webots/sim/compose.gpu.yaml
+++ b/examples/webots/sim/compose.gpu.yaml
@@ -6,6 +6,12 @@
 
 services:
   sim:
+    environment:
+      # PRIME render-offload: prefer the passed-in dGPU for GL. On Optimus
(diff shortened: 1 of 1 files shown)


enkerewpo (approved):
LGTM

PR #109, opened 2026-06-28
feat: add soma subsystem

enkerewpo:
add support for soma.
previous PR: #107


system/soma/src/config.rs
```diff
@@ -0,0 +1,727 @@
...
+pub struct SomaConfig {
+    pub atlas_endpoint: String,
+    pub listen: String,
+    pub provider_id: String,
+    pub robonix_root: PathBuf,
+    pub default_robot: Option<String>,
+    pub deployments: Vec<DeploymentConfig>,
+    pub start_packages: bool,
+    pub rbnx_bin: String,
+}
```
enkerewpo: is `robonix_root`, `default_robot`, `deployments`, `start_packages`, `rbnx_bin` necessary? I prefer removing these fields. and why there's no `robot_yaml` field?
```yaml
soma:
  listen: 127.0.0.1:50091
  log: info
  robot_yaml: ./robot.yaml   # soma spec (the components tree), relative to the deploy dir
```

system/soma/config.yaml
enkerewpo, line 1: please remove this file

system/soma/src/report.rs
```diff
@@ -0,0 +1,94 @@
...
+                )
+            })
+        })
+    }
+
+    /// Print a compact startup table to stderr so operators can inspect bring-up results.
+    pub fn print_to_terminal(&self) {
+        eprintln!();
+        eprintln!("========== Soma startup report ==========");
+        for deployment in &self.deployments {
```
enkerewpo: there would be only ONE deployment of robonix running

system/soma/src/store.rs
```diff
@@ -0,0 +1,190 @@
...
+use thiserror::Error;
+
+#[derive(Debug, Clone)]
+pub struct SomaStore {
+    robots: BTreeMap<String, RobotRecord>,
+    default_robot: Option<String>,
+}
+
+#[derive(Debug, Clone)]
+pub struct RobotRecord {
```
enkerewpo: What are we having a RobotRecord and "Robot library" for? Each deploy folder is exactly a "robot" with all drivers/skills/robot description. Currently we don't save the locations of the deploy folder now, and these locations can by anywhere on disk.

system/atlas/proto/atlas.proto
```diff
@@ -531,6 +531,71 @@ message ListContractsResponse { repeated ContractDescriptor contracts = 1; }
...
+// -- Reverse notifications (provider event stream) ---------------------------
+//
+// Atlas has historically been pull-only: providers push state via
+// SetLifecycleState / Heartbeat, consumers pull via Query / ConnectCapability.
+// The boot orchestrator (Soma) needs a reverse path — rbnx finishes
+// bringing up the rest of the system services and then has to tell Soma
+// "ok, go run your stage-2 skill bring-up now". This pair of RPCs gives
+// any registered provider a way to subscribe to events targeted at its
+// own provider_id, and gives anyone with an Atlas channel a way to send
+// such events.
```
enkerewpo: changing atlas's core proto interface is very dangerous, why are we adding these new methods for soma? can it be used in a broader range? if not, we should not touch atlas for the interaction between components

capabilities/system/soma/get_urdf.v1.toml
```diff
@@ -0,0 +1,9 @@
+# Return the URDF referenced by a Soma YAML file.
+[contract]
+id      = "robonix/system/soma/get_urdf"
+version = "1"
+kind    = "system"
```
enkerewpo: should be kind="service"

capabilities/system/soma/get_yaml.v1.toml
```diff
@@ -0,0 +1,9 @@
+# Return raw Soma YAML for Pilot prompt construction.
+[contract]
+id      = "robonix/system/soma/get_yaml"
+version = "1"
+kind    = "system"
```
enkerewpo: should be kind="service"

system/atlas/proto/atlas.proto
lhw2002426: Now soma start has 2 stage, stage2 is after altas start, If there is not a iinterface to notify soma, How can soma stage2 know atlas start finish?

examples/webots/soma.yaml
```diff
@@ -0,0 +1,83 @@
...
+  passable_door_width_m: 0.80
+  exports:
+    - provider_id: tiago_webots_nav2
+      capabilities:
+        - { path: robonix/service/navigation/navigate, description: "Navigate to a 2D goal in the mapped scene." }
+        - { path: robonix/service/navigation/navigate/status, description: "Report the current navigation goal status." }
+        - { path: robonix/service/navigation/navigate/cancel, description: "Cancel the active navigation goal." }
+    - provider_id: skill_explore_room
+      capabilities:
+        - { path: robonix/skill/explore/room, description: "Run autonomous frontier exploration when the explore skill package is available." }
```
enkerewpo: the explore skill does not have a `room` capability, please check <https://github.com/enkerewpo/explore_rbnx/blob/main/package_manifest.yaml>

system/soma/README.md
```diff
@@ -1,25 +1,238 @@
...
+
+```yaml
+exports:
+  - provider_id: tiago_webots_chassis
+    capabilities:
+      - { path: robonix/primitive/chassis/move, description: "Command chassis motion." }
+      - { path: robonix/primitive/chassis/odom, description: "Read odometry." }
+  - provider_id: skill_explore_room
+    capabilities:
+      - { path: robonix/skill/explore/room, description: "Explore the current room." }
```
enkerewpo, line 176: same, no `explore/room`

tools/rbnx/src/cmd/deploy.rs
```diff
@@ -590,6 +819,41 @@ pub async fn execute(
         expand_yaml(&mut e.config);
     }
 
+    // soma owns primitive + skill bring-up (see
+    // docs/soma_two_stage_bringup.md). If the manifest declares ANY
```
enkerewpo, line 884: the doc does not exist, you can later create a PR to `robonix-book` for the soma part

enkerewpo:
@robonix-ci test

enkerewpo:
CI reached the real Webots boot path and failed before scenarios.

Run: https://github.com/syswonder/robonix/actions/runs/28586047282  
Report: https://syswonder.github.io/robonix/reports/runs/28586047282/

Root cause from the uploaded logs:

```text
soma.log:
Error: resolve Soma config
Caused by:
    missing robot_yaml: set --robot-yaml, ROBONIX_SOMA_ROBOT_YAML, or provide it in --config <yaml>

rbnx-boot.log:
[FAIL] soma stage 1 timeout after 180s; service bring-up needs primitives ACTIVE first
Error: soma stage 1 did not become ACTIVE within 180s; refusing to start service packages before primitives are ready
```

So the failure is not the CI bot, runner, Webots simulator, or scenario harness. The simulator and fake VLM started; `rbnx boot` failed because `robonix-soma` started without a `robot_yaml` path and exited, then `rbnx` waited 180s for Soma stage 1 / primitive ACTIVE state.

The merge result contains `examples/webots/soma.yaml`, but `examples/webots/robonix_manifest.yaml` does not provide `system.soma.robot_yaml`. In `tools/rbnx/src/cmd/deploy.rs`, `ensure_soma_defaults()` creates/promotes `system.soma`, but it explicitly does not inject a default `robot_yaml`. That means an auto-implied Soma is spawned with no `--robot-yaml`, which cannot work.

Suggested functional fix:

1. Make the Webots deploy pass Soma a robot YAML, e.g. add an explicit `system.soma` block with `robot_yaml: ./soma.yaml` and make sure `rbnx boot` resolves that path relative to the manifest dir.
2. Or, if the intended convention is `soma.yaml` next to the manifest, make `ensure_soma_defaults()` inject `manifest_dir/soma.yaml` when primitives/skills imply Soma and that file exists. If it does not exist, fail before spawning Soma with a direct manifest/config error.
3. Improve boot failure reporting: while waiting for `soma stage 1`, `rbnx` should notice that the Soma child exited and surface `soma.log` / the child exit instead of waiting the full 180s timeout.

Until this is fixed, scenarios cannot run because primitives are delegated to Soma stage 1 and never become ACTIVE.

After you fixed it, @ me to trigger the webots CI

enkerewpo:
@robonix-ci test

enkerewpo:
CI did not reach the scenario suite for this run. This is a Soma stage-2 bug, not a Webots scenario failure.

Artifact / report links:

- Report: https://syswonder.github.io/robonix/reports/runs/28701732472/
- Action run: https://github.com/syswonder/robonix/actions/runs/28701732472
- Raw log artifact: https://github.com/syswonder/robonix/actions/runs/28701732472/artifacts/8080537807

What happened, based on the logs:

- The workflow failed at `Boot Robonix deployment`.
- `Run interface checks`, `Wait for mapping occupancy grid`, and `Run scenario suite` were skipped.
- This is why the HTML report shows `NO DATA` / `0/0 scenarios`: the scenario runner never executed.

The important boot log evidence is in `examples/webots/rbnx-boot/logs/soma.log`:

```text
17:30:59.052145 soma: stage 2 watcher armed on fd 3; waiting for 'stage2' line
17:30:59.052774 soma: stage 2 (skill bring-up) failed: read stage-trigger pipe: Resource temporarily unavailable (os error 11)
```

Later, `rbnx-boot.log` prints:

```text
[ OK ] soma stage 2 trigger delivered
```

but that only means the rbnx side wrote `stage2\n` to the pipe. It does not prove that Soma was still reading the pipe, and it does not prove that skill bring-up ran. The final caps snapshot confirms this: `soma`, primitives, `mapping`, `nav2`, `scene`, etc. are ACTIVE, but there is no `explore` provider and no `robonix/skill/explore/*` capabilities. The workflow requires those caps, so it waited until the boot polling timeout and then failed before scenarios.

The code-level cause is in `system/soma/src/main.rs`:

```rust
set_nonblocking(&file).context("mark stage-trigger fd non-blocking")?;
let async_file = tokio::fs::File::from_std(file);
let mut reader = BufReader::new(async_file);
...
let n = reader
    .read_line(&mut line)
    .await
    .context("read stage-trigger pipe")?;
```

This converts the inherited pipe fd to `O_NONBLOCK`, then immediately calls `read_line()`. At that point rbnx has not written the stage-2 trigger yet, because rbnx only writes it after the rest of system/service bring-up. For a nonblocking pipe with no data currently available, Linux returns `EAGAIN` / `WouldBlock`. The current code treats that as a fatal read error for stage 2, logs a warning, and exits the watcher. That creates a race: Soma can arm the watcher, hit `EAGAIN` before rbnx writes, and permanently skip skill bring-up.

This matches the observed run exactly: Soma stage 1 and core services came up, but stage 2 exited early, so `explore` never registered.

Suggested fix:

1. Do not treat `EAGAIN` / `WouldBlock` as a stage-2 failure. The watcher must continue waiting until it reads `stage2\n`, sees EOF, or receives shutdown.
2. Prefer one of these implementations:
   - Keep the pipe fd blocking and read the single line inside `tokio::task::spawn_blocking`, then run skill bring-up after the blocking read returns.
   - Or use `tokio::io::unix::AsyncFd` correctly: wait for readable readiness, attempt the read, and if the read returns `WouldBlock`, clear readiness and wait again.
3. Make stage-2 failure observable to rbnx/CI. In CI, skills are not optional: `explore` is required by the Webots integration scenarios. If Soma cannot run stage 2, boot should fail with a clear reason rather than only logging a warning inside `soma.log`.
4. Change the rbnx message from `stage 2 trigger delivered` to something less misleading, e.g. `stage 2 trigger written`, unless rbnx also waits for the expected skill providers/caps to register.
5. Add a regression test for the delayed-writer case: start `run_stage2()` with a pipe whose writer sends `stage2\n` after a delay. The test should verify that the watcher does not exit on the initial empty read and that `spawn_skills()` is reached.

References:

- Linux `read(2)`: with `O_NONBLOCK`, a read that would block may fail with `EAGAIN` / `EWOULDBLOCK`: https://man7.org/linux/man-pages/man2/read.2.html
- Linux `pipe(7)`: if a process attempts to read from an empty pipe and `O_NONBLOCK` is enabled, `read(2)` fails with `EAGAIN` when writers still exist: https://man7.org/linux/man-pages/man7/pipe.7.html
- Tokio `AsyncFd`: the async pattern is to wait for fd readiness and retry on `WouldBlock`, not to treat the first `WouldBlock` as terminal: https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html

lhw2002426:
@robonix-ci test

lhw2002426:
@robonix-ci test

enkerewpo:
LGTM

PR #116, opened 2026-06-29
feat(scene): VLM-primary scene graph + stable object identity 

HeartLinked:
# Summary

Two independent changes, both verified against the latest dev branch:

1. `feat(scene)` — Rework the scene-graph relation subsystem to be VLM-primary and keep object identity stable across perception ticks.
2. `fix(pilot)` — Tolerate prose around the RTDL JSON reply and tighten the envelope prompt.

## Scene: VLM-primary scene graph + stable object identity

Relations are now VLM-primary. The store holds two disjoint writer slices, composed on read for the MCP tools and `/api/state`:

* **Fast geometric loop**
  `geometric_loop.py`, ~3 Hz, always on, including when VLM credentials are absent.
  Owns only `reachable_by`, using a real 3D gripper-to-object distance. Contact and containment have moved out of this path because the old AABB tests misfired on full-volume boxes and same-surface objects.

* **Image-grounded VLM**
  `builder.py` + `image_relations.py`, ~30 s, gated by `SCENE_GRAPH_ENABLED`.
  Projects tracked objects into the current RGB keyframe, draws numbered boxes, and enumerates relational and semantic edges in one multimodal call. Falls back to the legacy text-only per-pair inference path when a camera frame bundle is unavailable.

* **Legacy relation engine removed**
  The old text-only `state/relations.py` `RelationEngine` has been removed.

Object identity is stabilized. Incremental `MapObjectList` → `ObjectRegistry` reconciliation with UUID rebind + soft-evict keeps an object’s ID and `observation_count` stable across concept-graphs UUID churn. This fixes the “N objects collapse to 1” flicker. The change also adds same-class proximity and class-agnostic geometric collapse passes.

## Config

Documented in `system/scene/README.md`:

* New `SCENE_CG_*` merge/dedup tuning knobs.
* New `SCENE_OBJECT_TTL_SEC` tuning knob.
* These settings are forwarded into the container.
* `VLM_REASONING_EFFORT` is opt-in:

  * Unset → the field is omitted, so non-reasoning models and strict OpenAI-compatible endpoints remain unaffected.
  * Set to `minimal` to keep a reasoning `VLM_MODEL` fast.

## Tests

Ships model-free unit coverage:

* `test_identity.py`
* `test_image_relations.py`
* Extended `test_scene_graph.py`

## Pilot: tolerate prose around RTDL JSON

`parse_rtdl_assistant_response` now extracts the first balanced top-level `{…}` before parsing, using string-aware brace matching.

This allows replies with a narrated line before the JSON or a trailing comment after the JSON to parse successfully.

The protocol and retry prompts are tightened:

* The reply must be a bare JSON object.
* Every value must be a literal.
* Expressions such as `Math.sin(...)` are rejected.
* `task_update` must be `null` or a full object.
* `task_update.status` must be `"in_progress"` or `"done"`.

Two new unit tests cover:

* Preamble stripping.
* Braces inside strings.

## Validation

* Rust, full workspace, build host:

  * `cargo fmt --all -- --check` ✅
  * `cargo clippy --workspace --all-targets -- -D warnings` ✅
  * `cargo test --workspace` ✅
  * Pilot: 23/23 tests passing, including the 2 new tests.
* `scripts/check-docs.sh` ✅

  * English-only check passed.
* Scene ran end-to-end in the Webots stack:

  * Perception → geometric loop + image-grounded relations → MCP/web.


enkerewpo:
please summarize your PR in description

system/pilot/src/planner.rs
```diff
@@ -1226,7 +1279,9 @@ fn build_rtdl_retry_prompt(
          Previous response preview: {}\n\n\
          Fix the RTDL error and retry the same user request exactly once. If the error \
          mentions an unknown capability, do not repeat that capability. Return ONLY a JSON object \
-         with exactly `content`, `rtdl_description`, `rtdl`, and `task_update`. Use only \
+         with exactly `content`, `rtdl_description`, `rtdl`, and `task_update`. The reply MUST begin \
+         with `{{` and end with `}}`: no prose, narration, or markdown fences before or after it (put \
```
enkerewpo, line 1300: why `{{` and `}}`

HeartLinked: This string is built with Rust's format! macro (note the Previous response preview: {} placeholder that gets filled by raw_preview(raw_content)), and inside a format! template the characters { and } are reserved for argument placeholders. To emit a literal brace you therefore have to escape it by doubling it: {{ renders as a single { and }} renders as a single }. So when the model actually receives the prompt, it reads …must begin with `{` and end with `}`…— the doubling is purely Rust format-string escaping, not part of the text. (By contrast, the RTDL_PROTOCOL_REMINDER constant is a plain&strliteral, not aformat!t emplate, so it correctly uses single{and}` directly.)

HeartLinked:
> please summarize your PR in description

done

system/pilot/src/planner.rs
enkerewpo: OK

enkerewpo:
please resolve the conflicts so we can test this PR in webots CI

enkerewpo:
@robonix-ci test

enkerewpo:
I'm fixing the CI, after that I will re-issue a webots integration test

enkerewpo:
@robonix-ci test

enkerewpo (approved):

enkerewpo:
LGTM

PR #119, opened 2026-07-03
feat(capabilities): add hand primitive contracts and IDL for dexterous hand

Ciliphen:
## Summary

  Introduce `robonix/primitive/hand/*` as a new standard primitive kind for dexterous hands.

  ## New capability contracts (7)

  | contract_id | mode | MCP | description |
  |-------------|------|:---:|-------------|
  | `robonix/primitive/hand/driver` | rpc | — | lifecycle entry point |
  | `robonix/primitive/hand/move_joint` | rpc | ✅ | set ND joint positions (0=open, 1=closed) |
  | `robonix/primitive/hand/state_joint` | topic_out | — | continuous joint position stream |
  | `robonix/primitive/hand/state_speed` | topic_out | — | continuous joint speed stream |
  | `robonix/primitive/hand/set_speed` | rpc | — | set speed limits per joint |
  | `robonix/primitive/hand/set_torque` | rpc | — | set torque limits per joint |
  | `robonix/primitive/hand/info` | rpc | ✅ | query joint names/count/model |

  ## New IDL (5)

  Under `capabilities/lib/hand/`:
  - `srv/MoveJoint.srv` — `float64[] positions → bool ok, string message`
  - `srv/SetSpeed.srv` — `float64[] speed → bool ok`
  - `srv/SetTorque.srv` — `float64[] torque → bool ok`
  - `srv/GetHandInfo.srv` — `→ string[] joint_names, string model, uint32 num_joints`
  - `msg/JointState.msg` — `float64[] positions`

  ## Design

  - Control/feedback symmetry: `move_joint` ↔ `state_joint`
  - Control interfaces exposed as MCP for LLM consumption; high-frequency feedback via topic_out
  - Speed/torque parameters NOT exposed as MCP (set by skill/service in task context)
  - Model-agnostic: consumer queries `info` first to discover joint count, then adapts `move_joint` array length dynamically

  ## Verification

  ```bash
  # Check TOML format
  for f in capabilities/primitive/hand/*.toml; do echo "--- $f"; cat "$f"; done

  # List IDL files
  find capabilities/lib/hand -type f | sort

  References

  - Reference implementation: BeingBeyond D1 DexHand (Beingbeyond_D1/primitives/d1_hand/)

  Impact

  - No breaking changes
  - Adds hand to the primitive kind set alongside chassis/camera/lidar/imu/audio


enkerewpo:
maybe 
`robonix/primitive/hand/set_speed_limits` and 
`robonix/primitive/hand/set_torque_limits`?

enkerewpo (changes requested):
Please resolve copilot’s review comments

capabilities/lib/hand/srv/MoveJoint.srv
```diff
@@ -0,0 +1,4 @@
+float64[] positions
```
enkerewpo: A bare `float64[] positions` makes joint order an implicit ABI between `info` and `move_joint`. For dexterous hands from different vendors this is fragile: callers can send the right length in the wrong order and the service cannot detect it. Please include `string[] joint_names` alongside `positions`, or make the joint order a named schema field returned by `info` and validated by the driver.

capabilities/lib/hand/srv/SetSpeedLimits.srv
```diff
@@ -0,0 +1,3 @@
+float64[] limits
```
enkerewpo: For set xxx limits, should we also add “names” like the others?

capabilities/lib/hand/srv/SetTorqueLimits.srv
```diff
@@ -0,0 +1,3 @@
+float64[] limits
```
enkerewpo: same, maybe we should add “names” as well

enkerewpo:
@robonix-ci test

capabilities/lib/hand/srv/SetTorqueLimits.srv
Ciliphen: Noted. This push was triggered due to my operational error, and I am still modifying the code at present.

enkerewpo:
Thanks for adding the hand primitive contracts!

I think the interface still needs clearer semantics before merge; otherwise different hand drivers may implement incompatible behavior.

Main points:

1. Units / value convention are not defined in the IDL. `positions` and `limits` are plain `float64[]`, while the TOML comments imply `0-1` normalized values. Please make this explicit in the `.srv/.msg` files too, or rename fields to something like `normalized_positions`, `normalized_speed_limits`, `normalized_torque_limits`.

2. `move_joint` execution semantics are unclear. Does `ok=true` mean the command was accepted, queued, or the hand has already reached the target position?

3. `joint_names + values` array semantics need to be specified. Are partial joint subsets allowed, or must every request include all joints from `info.joint_names`? Are duplicate names rejected?

4. `GetHandInfo` may be too small for safe model-agnostic control. If the public interface uses normalized values, we likely need per-joint range/direction metadata, or at least a documented convention for open/closed mapping.

5. `SetSpeedLimits` and `SetTorqueLimits` should probably return `string message` like `MoveJoint`, so rejected limit changes can explain length mismatch, unknown joint, out-of-range value, hardware rejection, etc.

Also, contract TOML should avoid saying “LLM-facing MCP tool”. The contract defines the abstract interface; whether a provider exposes it to the model/MCP should be decided by provider registration and capability description.

Ciliphen:
The current code abstracts movements at the finger level while retaining fine-grained control over the axes; for each finger, directions such as “bend,” “sway,” and “oppose” are defined and normalized on a 0–1 scale.

capabilities/lib/hand/msg/JointState.msg
lhw2002426, line 1: why not use sensor_msgs/msg/JointState.msg

Ciliphen: In ROS, JointState uses unit-bearing values such as radians. The dexterous hand I am currently referencing is the Linker Hand O6, and all parameters passed in its SDK adopt normalized values ranging from 0 to 1.

lhw2002426 (approved):
I have no problem about this

enkerewpo:
OK

enkerewpo (approved):

PR #121, opened 2026-07-04
feat(memory): extend memory with causal knowledge graph model

ohhhHwH:
Summary

Introduces Memgraph, a structured Causal Knowledge Graph (CKG) memory service for Robonix. Runs as an MCP service under robonix/service/memgraph/* in parallel with the existing
memsearch backend.

Embodied agents need multi-modal, spatially-aware, causally-linked, executable memory. Memgraph addresses three layers:

1. Remember — what happened (spatial + behavioural + cognitive + optimisation tags)
2. Retrieve — find relevant past (tag filter → BM25+Embedding hybrid → causal filter)
3. Refine — turn experience into skills (Phase 2, scaffolded)

afc4c50372aff4220c5e364e6d81a2fa
850951f28b3dcc5c1d7df1d99c162f7c
8ce1478e0a1ebee023899af599b6cc7c


.claude/settings.json
enkerewpo, line 1: please remove this file

ohhhHwH: Oops, I forgot

capabilities/service/memgraph/compact.v1.toml
```diff
@@ -0,0 +1,8 @@
+[contract]
+id      = "robonix/service/memgraph/compact"
```
enkerewpo: I suggest that we put everything (including the mem graph and the legacy memory) under `robonix/service/memory` namespace instead of separating them

examples/webots/robonix_manifest.yaml
```diff
@@ -86,6 +86,15 @@ primitive:
 
 # after booted all the primitives, start the services
 service:
+  # ── Memory services (two parallel backends) ──────────────────────────
+  # Memgraph (Scribe Mem) — structured CKG memory (BM25 + Embedding).
+  # Contracts: robonix/service/memgraph/{remember,search,compact}
+  # Embedding: all-MiniLM-L6-v2 (d=384), BM25 + Cosine hybrid retrieval.
+  - name: memgraph
```
enkerewpo, line 98: please use the name `memory` like before

capabilities/service/memgraph/compact.v1.toml
ohhhHwH: Got it — I'll update the plan to consolidate everything under robonix/service/memory, including both the mem graph and legacy memory.

enkerewpo:
@robonix-ci test

services/memory/memory_service/storage/graph_store.py
```diff
@@ -0,0 +1,259 @@
...
+
+        self.remove_node(node_id)
+
+        node.node_id = new_id
+        node.node_type = NodeType.LONG_TERM
+        self._nodes[new_id] = node
+        self._children[new_id] = set()
+        self._parents[new_id] = set()
+
+        for pid in old_parents:
```
enkerewpo, line 209: `MemoryNode.causal_chain` is the serialized copy of parent node ids (see `MemoryNode.causal_chain` in `core/types.py`), and `add_edge()` explicitly appends `parent_id` into that field on the child node.

The problem here is that `promote_to_long_term()` remaps the in-memory adjacency sets from `old_id` to `new_id`, but it does not rewrite the corresponding `causal_chain` entries stored inside affected child `MemoryNode`s.

So after compaction, these two representations can diverge:
- graph structure says `new_id -> child`
- serialized child node still says `causal_chain=[old_id]`

That leaves stale parent ids in persisted node data after compaction. Please rewrite affected child `causal_chain` values when remapping ids, and add a test that verifies both adjacency and serialized `causal_chain` stay consistent across promotion.

enkerewpo:
Thanks for your contribution. This is a pretty interesting direction. I do think a richer structured memory backend is worth doing, especially if we want memory to hold more than simple text notes.

I still have a few design questions though, and I think it would help to make them explicit in this PR.

Right now my understanding is:
- `memsearch` is still the current simple memory implementation, and `pilot` still depends on it
- this new graph-based memory is meant to provide richer structured memory
- for now the two may coexist
- later we may want to converge them under one `memory` abstraction

If that is the plan, I think the PR should say it more clearly.

At the moment I still cannot quite tell:
- whether this is supposed to replace `memsearch` now, or just live alongside it
- what exactly should still go to `memsearch`
- what should go to this new graph memory
- who is expected to write into it
- whether the two stores are intentionally separate for now, or supposed to stay in sync somehow

If the intended model is basically “`memsearch` stays as the simple memo-style memory for now, and this graph backend adds a richer structured path”, then I think stating that directly would make the design much easier to understand and review.

ohhhHwH:
> Yes — that is exactly the intended model, and thanks for calling it out.
> Let me explain the design rationale more explicitly.
>
> **memsearch** is a *declarative* note-pad: Pilot explicitly calls
> `save(“remember this fact”)` when the LLM decides something is worth
> keeping.  The stored content is free-form text — a user preference, a
> one-off observation, a reminder.  It stays as the default lightweight
> path for memo-style persistence.
>
> **memgraph** is an *observational* knowledge graph: it is **not** driven
> by Pilot's explicit save calls.  Instead, the robot's own execution
> pipeline — Executor (action results), Scene (perceived objects and their
> 3D positions), and downstream Scribe Log feeds — writes structured
> `MemoryNode`s that record **what the robot did, what it perceived, where
> it was, and how those events are causally linked**.  Each node carries:
>
> - **spatial tags** — scene type, object IDs + 3D coordinates
> - **behavioural tags** — action type (grasp / navigate / craft), success/failure, tool used
> - **cognitive tags** — task type, difficulty, intent
> - **causal edges** — “I navigated to the kitchen → I grasped the cup”
> - **embedding** — dense vector for hybrid BM25+semantic retrieval
>
> In short:
> | | memsearch | memgraph |
> |---|---|---|
> | **What is stored** | free-text facts & preferences | structured experience records |
> | **Who writes** | Pilot (LLM decides) | Executor + Scene + Scribe (automatic) |
> | **What is retrievable** | “what did the user say?” | “where was the cup? how did I grasp it? what step came before?” |
> | **Query dimensions** | text only | spatial + behavioural + cognitive + temporal + semantic |
>
> They are **intentionally separate** and **do not need to stay in sync** —
> they serve different layers of memory.  memsearch answers *”what did the
> user tell me?”*; memgraph answers *”what happened, where, why, and what
> did I learn from it?”*.  The latter is the foundation for skill refinement
> (N successful experiences → extract a reusable Skill template) and
> failure-lesson protection (failed actions are guarded from forgetting).

ohhhHwH:
> Admittedly, in the **current Phase 1 implementation** there is some
> functional overlap: both memsearch's `search` and memgraph's
> `hybrid_search` appear to be "input query text → return relevant
> memories", both registered as MCP tools for Pilot/LLM consumption.
> This is a deliberate short-term compromise — while memgraph's causal
> filtering, executability checking, and skill refinement are still
> under development, both backends need a working retrieval entry point.

ohhhHwH:
@robonix-ci test

enkerewpo:
@robonix-ci test

enkerewpo:
> ### Robonix Webots CI - not authorized
> @ohhhHwH cannot start this self-hosted Webots test run. Required repo permission: `admin` or `maintain`. Current permission: `write`.

actually I've added you as maintainer before, this might be a ci check bug, I will investigate it later

ohhhHwH:
@robonix-ci test

enkerewpo:
@robonix-ci test

enkerewpo:
Robonix will undergo a repository history cleanup. To keep this work available while avoiding further development on fork history that will soon be rewritten, we will preserve the current PR head exactly in the main repository as `feat/memory-causal-knowledge-graph` at `322b0104bdabc0a5b1497f7c9d33bb70b64a935a`.

This PR will then be closed without merging it into `dev`. Closing it is an administrative migration, not an acceptance or rejection of the implementation. The existing commits and authorship will remain intact on the preserved branch.

After the history cleanup, the contributor will be added to `syswonder/robonix` and can continue directly on `feat/memory-causal-knowledge-graph`. Once the work is ready and reviewed, please open a new PR from that branch to `dev`.

PR #132, opened 2026-07-07
FEAT: add more webots scenarios

1mujue:
Originally, we only have one scenario(office.wbt) in examples/webots, which is not enough to prove generalization of Robonix, therefore, I add some new scenarios(kitchen.wbt, appartment.wbt, complte_apartment.wbt, break_room.wbt), and now you can choose the world when start the simluation, like:
`bash sim/start.sh --world kitchen.wbt`.
If you didn't pass in --world arg, then it will start office.wbt as default.


enkerewpo:
@robonix-ci test

enkerewpo (approved):
LGTM, thanks!

PR #134, opened 2026-07-08
fix(runtime): clean package shutdown and restart

### LINKED ISSUE #128: [BUG] Ctrl-C one `rbnx boot`, and the next one is guaranteed to fail
## What happens

When `rbnx boot` exits early — for example via Ctrl-C, or because a stage fails mid-boot — it leaves a bunch of things behind.Three layers survive:

1. Host-side service processes keep running:

   * `memsearch`
   * `speech`
   * `voiceprint`
   * bridges

2. Service containers also keep running:

   * `scene`
   * `mapping`
   * `nav2`
   * `explore`

   The `rbnx start` wrapper dies as a zombie, so the `docker stop` trap in `start.sh` never gets a chance to run.

3. The sneakiest part: the chassis / camera / lidar drivers that soma `docker exec`s into the sim container outlive soma itself.

On top of that, entries in `~/.robonix/processes.json` are left behind.

The next boot is then basically doomed. Soma tries to spawn each primitive, but every spawn hits:

```text
Process for package::X already running, skipping
```

Then nothing registers with atlas, soma declares stage 1 dead, and exits.

So the pattern is very consistent:

```text
Ctrl-C once → next boot fails every time
```

The only way out right now is to manually clean everything up: pkill the leftover processes, stop the containers, kill the in-sim drivers, and reset the process registry JSON.One extra painful bit: the real failure reason only goes to stdout. `soma.log` ends after a few `Stopping process` lines and has zero `ERROR` lines, so it is hard to tell what actually failed.

## Repro

1. Run:

   ```bash
   cd examples/webots
   rbnx boot
   ```

2. Wait until it reaches `ACTIVE`.

3. Press Ctrl-C.

4. Run again:

   ```bash
   rbnx boot
   ```

5. The second boot fails with:

   ```text
   [FAIL] soma stage 1 — soma exited before becoming ACTIVE
   ```

## Expected behavior

Aborting boot should tear down everything it started, including:

* host-side service processes
* service containers
* drivers launched into the sim container via `docker exec`
* stale entries in `~/.robonix/processes.json`

Soma should also be more tolerant of leftovers. When it sees an already-running primitive, it should either adopt it or kill and respawn it. Both sides matter: clean teardown handles normal exits, while tolerant startup handles `kill -9`, power loss, crashes, and other messy cases.

When stage 1 fails, `soma.log` should also say which primitive failed and why.

enkerewpo:
Fixes #128.

This PR has two commits that together close issue #128 and add CI coverage so the regression can never ship again.

---

## Commit 1: `4dd1e01c` — fix(runtime): stop soma package children cleanly

Unifies package lifecycle teardown so `rbnx boot` and Soma use the same shutdown path:

- call `Driver(CMD_SHUTDOWN)` when a driver is known
- run optional manifest-level stop hooks
- TERM/KILL the wrapper process group
- prune stale `ProcessManager` state by checking zombie state, command line, and package cwd instead of trusting `kill(pid, 0)`

It also fixes the Webots Tiago chassis primitive PYTHONPATH so the container can import generated protobuf modules on a fresh boot.

Related external package stop-hook PRs:
- syswonder/service-map-rbnx#1
- syswonder/service-navigation-rbnx#1
- syswonder/skill-explore-rbnx#1

Validation on workpc Debian 13 / apartment Webots:
- `cargo test -p robonix-cli -p robonix-soma`
- full `rbnx boot` reached 11 components up
- `rbnx shutdown` completed in ~9s with only `robonix_tiago_sim` left running and `rbnx-boot/state.json` removed
- second full `rbnx boot` reached 11 components up again
- speech service loaded FunASR streaming ASR on CUDA and became ACTIVE

---

## Commit 2: `e1825138` — ci(testing): add boot/shutdown lifecycle validator; runtime: per-deploy process state

The first commit fixed the bug, but the Webots CI still only exercised `rbnx boot` once. So issue #128 could reappear in any future change without CI catching it. This commit closes that gap and addresses the follow-up "process state should be in rbnx-boot, not ~/.robonix".

### 2a. `testing/validate_lifecycle.py` + new CI step "Validate boot/shutdown lifecycle"

A new CI step inserted right after `Run scenario suite` and before `Capture runtime logs` runs the full lifecycle on a live deploy:

```
boot (already up) -> shutdown -> re-boot -> re-run scenarios -> shutdown (final)
```

The validator fails the workflow (exit non-zero) on any of:

- `<manifest-dir>/rbnx-boot/state.json` left behind after a shutdown
- leftover `rbnx boot` / `robonix-atlas` / `robonix-pilot` / `robonix-soma` / `robonix-liaison` processes
- the four per-run ports (atlas / executor / pilot / liaison) still bound
- zombie provider-owned Docker containers (other than the long-running sim container, which is reused for the second boot)

A new step "Re-run scenario suite after lifecycle validator" then re-runs the full scenario suite on the post-shutdown boot, so the report covers both a fresh boot and a clean-restart boot — i.e. the suite actually exercises the fix in commit 1 instead of just trusting it.

Lifecycle logs land in `<manifest-dir>/rbnx-boot/logs/lifecycle-*.log`; the new `examples/webots/rbnx-boot/processes.json` is also uploaded as a diagnostic when the validator fails. The existing `Stop run resources` step is now a no-op for normal runs (the validator's final shutdown already cleaned up), but stays as a safety net for runs that fail before reaching the validator.

### 2b. `tools/rbnx/src/process.rs` — per-deploy process state

`rbnx start` previously wrote its package-runtime state to `~/.robonix/processes.json`. That crossed the wrong boundary:

- `~/.robonix/` is for cross-deploy user state (`config.yaml`, `chat.yaml`, `voiceprint.json`, installed package DB).
- The package runtime record is per-deploy and must die with its manifest.
- `rbnx shutdown` only knows about the per-deploy `rbnx-boot/state.json`, so a leaked `~/.robonix/processes.json` from a half-killed boot was never cleaned up and could confuse `rbnx start` on unrelated packages on the next run.

The state file now lives next to its deploy:

- `<manifest-dir>/rbnx-boot/processes.json` when `rbnx boot` (or any caller that sets `SCRIBE_LOG_DIR=<…>/rbnx-boot/logs`) brought the wrapper up
- `<pkg>/rbnx-build/processes.json` for standalone `rbnx start` whose default log dir is `<pkg>/rbnx-build/logs`

Both sit next to the logs that already follow the same layout, so a single `rm -rf examples/webots/rbnx-boot` (or `rbnx clean`) now wipes the whole deploy state.

### Why a separate commit

The shutdown fix (commit 1) and the CI coverage + state-location fix (commit 2) are independent: the bug fix ships on its own, the CI coverage ships on its own. Reviewing them together would mix "does the fix work" with "do we test the fix works" — splitting lets the reviewer bisect either concern.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -920,6 +920,55 @@ jobs:
           # Surface the one-line RESULT for the reusable-workflow output / bot.
           echo "result=$(grep '^RESULT:' "$RUNNER_TEMP/scenarios.out" | tail -1)" >> "$GITHUB_OUTPUT"
 
+      - name: Validate boot/shutdown lifecycle
+        # boot → shutdown → re-boot → shutdown. Proves the runtime keeps
+        # state under <manifest-dir>/rbnx-boot/ (not ~/.robonix/), that
+        # Soma package children are stopped cleanly, and that the stack
+        # can come back up after a clean teardown (the issue #128
+        # regression). The second boot reuses the per-run sim container;
+        # the validator does NOT start a new webots instance.
+        working-directory: examples/webots
+        env:
+          ROBONIX_SIM_CONTAINER: "ci-${{ github.run_id }}-sim"
+        run: |
+          set -euo pipefail
+          python3 "$GITHUB_WORKSPACE/testing/validate_lifecycle.py" \
+            --rbnx rbnx \
+            --manifest robonix_manifest.ci.generated.yaml \
+            --server "$ROBONIX_CI_ATLAS" \
+            --sim-container "$ROBONIX_SIM_CONTAINER" \
+            --log-dir examples/webots/rbnx-boot/logs | tee "$RUNNER_TEMP/lifecycle.out"
+
+      - name: Re-run scenario suite after lifecycle validator
+        # After the validator re-boots the stack, run the scenarios one
+        # more time so the report covers both a fresh boot and a
+        # post-shutdown boot. This is what actually exercises the
+        # "restart doesn't crash" fix — the original suite proved
+        # boot-once, this proves boot-after-shutdown.
+        working-directory: examples/webots
+        run: |
+          set -euo pipefail
+          export ROBONIX_ATLAS="$ROBONIX_CI_ATLAS"
+          # Wait for atlas to settle (the validator just re-booted
+          # everything; reuse the same caps wait).
+          for i in $(seq 1 60); do
+            if rbnx caps -v --server "$ROBONIX_CI_ATLAS" >"$RUNNER_TEMP/caps.post-restart.txt" 2>&1 \
+              && grep -q "robonix/service/navigation/navigate" "$RUNNER_TEMP/caps.post-restart.txt" \
+              && grep -q "^● nav2 \[ACTIVE\]" "$RUNNER_TEMP/caps.post-restart.txt"; then
+              echo "post-restart atlas ready"
+              break
+            fi
+            if [ "$i" -eq 60 ]; then
+              echo "::error::atlas not ready after lifecycle reboot"
+              tail -120 "$RUNNER_TEMP/caps.post-restart.txt" || true
+              exit 1
+            fi
+            sleep 5
+          done
+          python3 "$GITHUB_WORKSPACE/testing/run.py" --server "$ROBONIX_CI_ATLAS" \
+            --summary-json "$GITHUB_WORKSPACE/testing/logs/summary.restart.json" \
+            | tee "$RUNNER_TEMP/scenarios.restart.out"
+
       - name: Capture runtime logs
         if: always()
         env:
@@ -1242,6 +1291,11 @@ jobs:
             ${{ runner.temp }}/rbnx-boot.log
             ${{ runner.temp }}/cache-provider-logs/
             ${{ runner.temp }}/sim-logs/
+            ${{ runner.temp }}/lifecycle.out
+            ${{ runner.temp }}/scenarios.out
+            ${{ runner.temp }}/scenarios.restart.out
+            examples/webots/rbnx-boot/logs/lifecycle-*.log
+            examples/webots/rbnx-boot/processes.json
           if-no-files-found: ignore
 
   report:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -921,12 +921,14 @@ jobs:
           echo "result=$(grep '^RESULT:' "$RUNNER_TEMP/scenarios.out" | tail -1)" >> "$GITHUB_OUTPUT"
 
       - name: Validate boot/shutdown lifecycle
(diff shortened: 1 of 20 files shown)


enkerewpo:
@robonix-ci test

enkerewpo:
> ### ❌ robonix CI — no scenario summary
> No machine-readable scenario summary was produced. Treat this as an incomplete infrastructure run, not as a verified Webots pass.
> 
> **LLM analysis** This PR adds explicit `stop:` hooks to all package manifests and a new CI lifecycle validator that tests boot → shutdown → re-boot. The Soma launcher and `rbnx boot`/`shutdown` commands are refactored to use a shared `shutdown_package_runtime` function that calls Driver(CMD_SHUTDOWN), then the manifest stop hook, then SIGTERM/SIGKILL. The CI workflow now runs the lifecycle validator after the scenario suite and then re-runs scenarios on the restarted stack. The report shows 'NO DATA' because no
> 
> Change summary: Added `stop:` hooks to package manifests for audio_driver, audio_macos_bridge, tiago_camera, tiago_lidar, tiago_chassis, simple_nav, memsearch, speech, voiceprint, and scene packages.; Added `stop` field to manifest schema in `tools/rbnx/src/manifest.rs`.; Refactored teardown in `tools/rbnx/src/cmd/teardown.rs` to use `shutdown_package_runtime` with atlas endpoint, stop hook, and process group termination.. Test result: CI built 11 packages successfully and booted the Webots stack. The lifecycle validator ran but its output is not captured in the logs. The scenario suite produced no machine-readable summary (0/0 scenarios). The Webots sim container logs show repeated 'extern controller' connection failures for my_robot and Ros2Supervisor, indicating the robot controller never successfully attached to Webots. The final caps check failed with 'Connection refused' to atlas, suggesting the stack was not running at Likely root cause: The Webots sim container's robot controller repeatedly failed to connect as an extern controller, preventing the simulation from starting. This is a pre-existing infrastructure issue unrelated to the PR changes. Suggested fix / watchout: Investigate why the Webots extern controller connection fails. Check that the robot name in the URDF matches the extern controller declaration, and that Webots is configured to accept extern controllers. This is likely a timing or configuration issue in the sim container entrypoint. Watchouts: The lifecycle validator and re-run scenarios may not have executed if the first boot failed; check CI logs for those steps.; The new stop hooks use pkill patterns that may be fragile if process names change.; The process state file relocation from ~/.robonix/ to per-deploy may break existing workflows that depend on the old path..
> 
> Report: [open HTML report](https://syswonder.github.io/robonix-ci-reports/reports/runs/28955029037/). Action run: [open GitHub Actions run](https://github.com/syswonder/robonix/actions/runs/28955029037). Raw logs: scenario JSONL, provider logs, simulator logs, final caps, and boot logs are in the same artifact.

since there'a issue on the webots ci tester server (with GPU), so the webots ci are not working now, I will fix it in the `dev` branch later

PR #137, opened 2026-07-09
feat(scene): save and restore rooms and objects with each SLAM map

enkerewpo:
## Problem

Saved SLAM maps could reload without their matching rooms and objects, and the first Load click sometimes only switched RTAB-Map while Scene still showed the previous binding. Save/Load also left the editor interactive, allowing duplicate requests or room edits during a map transition.

Pilot also passed labels such as `room 315` to an API that previously accepted only stable IDs.

## Design

- Mapping remains the owner of spatial artifacts; Scene stores rooms and objects by the same `map_id`.
- Load performs one RPC, waits for a newly published occupancy grid, then rebinds and restores Scene data. If the new grid is not observed, the request fails clearly instead of requiring another click.
- Saving an existing `map_id` preserves the immutable spatial artifact and updates its Scene rooms/objects.
- Save and Load now use a blocking operation modal. While active, map actions, room editing, drawing, pose estimation, and duplicate submissions are disabled.
- The modal shows the operation phases, keeps failure details visible, and provides explicit Retry/Close. Success is shown only after validation.
- Room goals support stable IDs, unique normalized names, and short aliases without fuzzy guessing.

## Tests

Environment: Jetson AGX Orin source/runtime dependencies; map RPCs were mocked for UI/API tests, so no real map, chassis, navigation, or arm action was triggered.

- Map operation and annotation Web/API tests: **20 passed**
- One-click Load regression verifies a fresh occupancy grid is observed before Scene rebind
- Embedded UI JavaScript: `node --check` passed
- Existing Scene suite: **80 passed**
- Earlier Jetson validation loaded a 200x118 map and restored 35 objects
- Stable room ID, `room 315`, and `315` resolved consistently

## Discussion

This PR remains Draft until the cold-start sequence is repeated on the real robot: save an unannotated partial map, restart/load, add rooms, save to the same `map_id`, then restart/load again.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1001,6 +1001,7 @@ jobs:
             --scene-url "http://127.0.0.1:${SCENE_WEB_PORT}" \
             --map-id "$map_id" \
             --mapping-container "${ROBONIX_MAPPING_CONTAINER:-robonix_mapping}" \
+            --maps-dir "$MAPPING_MAPS_DIR" \
             --sim-container "$ROBONIX_SIM_CONTAINER" \
             --delete-after \
             --timeout 420 | tee "$RUNNER_TEMP/sim-logs/scene-map-persistence.txt"
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -330,6 +330,7 @@ jobs:
       GRPCIO_PIN: "1.80.0"
       # ── Per-run isolation (coexist with interactive users on the shared box) ──
       RUN_ID: "${{ github.run_id }}"
+      ROBONIX_HOME: "${{ runner.temp }}/robonix-home"
       # GPU selection is runner-specific. Priority:
       # workflow_call input -> repo/org variable -> runner env -> default GPU 0.
       ROBONIX_CI_GPU_ID_INPUT: ${{ inputs.gpu_id || '' }}
@@ -988,6 +989,19 @@ jobs:
             --summary-json "$GITHUB_WORKSPACE/testing/logs/summary.json" | tee "$RUNNER_TEMP/scenarios.out"
           # Surface the one-line RESULT for the reusable-workflow output / bot.
           echo "result=$(grep '^RESULT:' "$RUNNER_TEMP/scenarios.out" | tail -1)" >> "$GITHUB_OUTPUT"
+
+      - name: Verify scene map persistence
+        run: |
+          set -euo pipefail
+          mkdir -p "$RUNNER_TEMP/sim-logs"
+          map_id="ci_scene_map_${RUN_ID}"
+          python3 "$GITHUB_WORKSPACE/testing/verify_scene_map_persistence.py" \
+            --scene-url "http://127.0.0.1:${SCENE_WEB_PORT}" \
+            --map-id "$map_id" \
+            --mapping-container "${ROBONIX_MAPPING_CONTAINER:-robonix_mapping}" \
+            --sim-container "$ROBONIX_SIM_CONTAINER" \
+            --delete-after \
+            --timeout 420 | tee "$RUNNER_TEMP/sim-logs/scene-map-persistence.txt"
 
       - name: Validate boot/shutdown lifecycle
         # boot → shutdown → re-boot. Proves the runtime keeps
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -330,7 +330,6 @@ jobs:
       GRPCIO_PIN: "1.80.0"
       # ── Per-run isolation (coexist with interactive users on the shared box) ──
       RUN_ID: "${{ github.run_id }}"
-      ROBONIX_HOME: "${{ runner.temp }}/robonix-home"
       # GPU selection is runner-specific. Priority:
       # workflow_call input -> repo/org variable -> runner env -> default GPU 0.
       ROBONIX_CI_GPU_ID_INPUT: ${{ inputs.gpu_id || '' }}
@@ -369,6 +368,9 @@ jobs:
             registry-1.docker.io:443
             auth.docker.io:443
             production.cloudflare.docker.com:443
+
+      - name: Configure runner-local Robonix home
+        run: echo "ROBONIX_HOME=$RUNNER_TEMP/robonix-home" >> "$GITHUB_ENV"
 
       - name: Select runner GPU
         run: |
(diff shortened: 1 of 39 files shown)


HeartLinked (commented):
Thanks for consolidating this. The transactional Load behavior—waiting for a fresh occupancy grid before rebinding—the symmetric Save/Load/Delete cleanup, and the CI persistence verifier are all solid additions.

I reviewed the full diff and found the following issues, roughly ordered by severity.

## High

### H1 — Object persistence still lacks epoch handling; Load restores mixed-epoch rows without filtering

The object schema in `persistence.py` has no generation field, and `load_all()` filters only by `map_id`. As a result, a partition written across multiple map rebuilds can restore stale-frame objects as mis-anchored markers, including markers outside the current map.

A simple reproduction is to load a legacy partition such as my `webots_lab` data: all historical mis-anchored objects return.

Setting `restore_on_start=false` hides the symptom during the default boot path, but the underlying mechanism remains unresolved.

I plan to address this in a follow-up by introducing generation-aware clearing for derived objects while retaining user annotations as user-owned assets. This does not need to block the current PR, but I am flagging it so we agree that epoch handling is still outstanding.

### H2 — After Save, the scene-graph builder continues persisting into the named partition

`_persist_scene_objects` in `web.py` rebinds the shared `object_store` to the target `map_id` and leaves it bound there. The builder in `service.py` holds the same instance and persists objects during every rebuild cycle.

In mapping mode, loop closures after Save can shift the frame. Those drifted coordinates then continue flowing into the named partition while the spatial artifact remains frozen at Save time under the `spatial_unchanged` behavior. The next Load can therefore restore objects against a different frame and mis-anchor them.

This also invalidates the PR’s assumption that the named partition remains equivalent to the snapshot captured at Save time.

### H3 — `.live-*` partitions accumulate without cleanup

Every boot creates a unique `.live-<pid>-<ts>` partition in `service.py`, and the builder persists rows into it during each cycle. There is no cleanup path, so the Milvus database grows continuously across restarts.

Annotations created before Save can also leave orphaned `.live-*.json` files.

Suggested approaches:

* Skip persistence while the store is bound to an ephemeral partition, and begin persisting only after Save or Load binds a named map.
* Sweep stale `.live-*` partitions and files during startup.

### H4 — `goal_room` blocks the event loop with unbounded synchronous work

`_occupancy_room_goal` calls `disc_inside_polygon` for every grid cell in the room’s bounding box. Each call performs 16 or more pure-Python point-in-polygon tests across the polygon vertices, and the entire operation runs synchronously inside `async def goal_room`.

For an 8 × 8 m room at 0.05 m resolution, this produces roughly 25,000 grid cells and more than 400,000 polygon tests. That can stall the entire scene service for several seconds, affecting the web API, MCP requests, and watchers.

At minimum, move this work to `asyncio.to_thread`. A stronger solution would rasterize the polygon with NumPy and avoid per-cell Python point-in-polygon checks.

## Medium

### M1 — `maps_save` can silently overwrite a saved map’s user annotations

`AnnotationStore.rebind(carry_current=True)` overwrites the target map’s JSON file with the current in-memory partition.

A problematic case occurs when the startup binding—through a lifecycle broadcast or `SCENE_MAP_ID`—already points to the target map, while `restore_on_start=false` means its annotations were never loaded. Saving that map then replaces previously saved rooms with the live session’s annotation set, which may be empty.

The `current_bound_id != map_id` conflict check does not cover this bound-but-not-restored state, so user-authored assets can be lost silently.

### M2 — The maps facade drops generation information

Both `maps_save` and `maps_load` call rebind with `generation=None`. This discards the epoch baseline used for stale-annotation detection.

After either facade operation, stale detection remains blind until the next lifecycle confirmation. The upstream mapping main currently does not broadcast that confirmation, so the baseline may remain absent indefinitely.

### M3 — Navigation does not surface stale annotations

`goal_room` does not check `room.stale` and can return a navigation goal inside a polygon that may have been re-anchored incorrectly.

`list_objects` also mixes stale rooms into the object list without distinguishing them from confirmed rooms. After a map rebuild, Pilot can therefore navigate to a stale annotation without any warning.

Suggested behavior:

* Refuse navigation to stale rooms and return a hint asking the user to confirm the annotation.
* Surface stale state explicitly in the object label or reason field.

### M4 — `SCENE_RESTORE_ON_START` is unreachable in Docker, and the README conflicts with current behavior

`SCENE_RESTORE_ON_START` is read only in `service.py`. `start.sh` does not pass it into the container, there is no manifest plumbing, and the README does not document it. In the Docker deployment, the setting is effectively unavailable.

The README also still describes warm restoration on boot in both the `SCENE_OBJECT_MEMORY_ENABLED` table row and the “Object memory (warm restore)” section, even though warm restore is no longer the default behavior.

This is the largest semantic change in the PR and currently has no user-facing documentation.

### M5 — Contract changes were made without regenerating reference documentation, and v1 contracts were mutated in place

`SaveMap.srv` removes the `database_path` response field. This is a breaking contract change, although mapping main has already been updated, so the current runtime remains aligned.

`GoalNear.srv` and `GetSceneGraph.srv` also change shape in place while retaining version `1`.

The `docs/` submodule pin remains unchanged. Per the repository guidelines, the following references should be regenerated with the project tooling after contract changes:

* `docs/src/reference/contracts.md`
* `docs/src/reference/idl.md`

## Also

### The rbnx README does not document `ROBONIX_HOME`

The bundled runtime commit adds the `ROBONIX_HOME` environment-variable override in `tools/rbnx/src/config.rs`, but `tools/rbnx/README.md` does not mention it.

The repository guidelines require crate READMEs to document important environment variables within the same contribution.

HeartLinked (approved):
Approving this now so it can land — the integration itself is solid (our side ran a full e2e on top of this branch: save → cold restart → Load ×3 with exact object/room restore, plus Reset → stale/flush behavior, all verified on the webots stack). The items below are known issues we agreed to defer to follow-up PRs rather than block this merge. Recording them here in lieu of separate issues:

@enkerewpo to fix in follow-up PRs:

- H4 — goal_room handler does blocking work on the event loop; long calls stall the scene web/MCP server.

- M3 — rooms marked stale are silently invisible to navigation (goal_room resolution); needs an explicit error/fallback so callers can tell "no such room" from "room is stale".

- M4 — SCENE_RESTORE_ON_START is not actually passed through to the service (manifest/launcher plumbing), and the README section describing it no longer matches the implemented behavior.

- M5 — generated reference docs were not regenerated for the new contracts (goal_room.v1.toml, SceneAnnotation.msg, etc.); regenerate via the codegen tooling rather than hand-editing.

- The remaining LOW-severity review notes from the PR discussion.

PR #140, opened 2026-07-11
fix(runtime): make planning state-aware and RTDL control reliable

enkerewpo:
## Problem

Pilot previously mixed task execution, cancellation, steering, and user-facing replies in one RTDL lifecycle. It also planned from stale local state instead of refreshing the authoritative Executor, Soma, and Scene snapshots. With concurrent or persistent plans this caused cancel loops, duplicate or delayed replies, incorrect running-task counts, and plans that ignored the robot's current motion, gripper, room, or nearby-object state.

## Design

- Add `ControlPlan` and `ListActivePlans` as an explicit Executor control plane.
- Keep cancel, cancel-all, and stop-boundary operations outside ordinary RTDL nodes.
- Target one plan without canceling independent plans or the control request itself.
- Terminate canceled Bash process groups instead of only dropping their streams.
- Refresh Pilot from Executor's authoritative active-plan snapshot before each planning round.
- Refresh embodiment state from Soma and environment state from Scene before planning.
- Add Scene's focused `get_robot_context` query so room, area, pose, and nearby objects do not depend on a truncated all-object listing.
- Scope steer replies to the newest input and emit one final reply at the actual terminal boundary.
- Preserve legacy Executor builtins for compatibility, while hiding them from the model.

## Scope

This PR contains only planning-context, RTDL execution/control, steering, cancellation, and reply-lifecycle changes. Shared lifecycle Driver migration, Vitals, Liaison/audio, Atlas namespace policy, and general `rbnx boot` cleanup remain separate work.

## Validation

After merging the current `dev` branch into the PR branch:

- `git diff --check`
- Executor unit tests: 36 passed
- Pilot unit tests: 47 passed
- Soma unit tests: 27 passed
- Scene robot-context test: 1 passed after isolated `rbnx codegen --mcp --ros2` against this PR checkout
- Total targeted tests: 111 passed, 0 failed

The existing branch also includes isolated Executor cancellation and natural-language RTDL integration scripts for the next full runtime validation. The PR remains draft until CI and the runtime interaction scenarios are reviewed.


system/pilot/src/service.rs
```diff
@@ -177,55 +290,148 @@ impl RobonixSystemPilot for PilotServiceImpl {
...
+                        true
+                    }
+                })
             } else {
                 false
             };
-            debug!("[pilot] abort_turn task for session {id} (signaled={ok})");
-            let (_tx, rx) = tokio::sync::mpsc::channel::<Result<PilotEvent, Status>>(1);
+            let executor_cancelled =
+                cancel_all_executor_plans(self.atlas.clone(), &self.provider_id).await;
```
enkerewpo, line 335: **[P1] `abort_turn` cancels plans owned by other sessions**

This sends a global `cancel_all` request, while `run_turn` already reacts to the watch signal by cancelling only the plans in the current supervisor forest. As a result, aborting one session can terminate plans from other sessions or plans started externally, which directly breaks the cancellation-isolation goal of this PR. Please remove the global cancellation or pass the current session's plan IDs through the control path.

system/pilot/src/planner.rs
```diff
@@ -842,13 +1360,50 @@ pub async fn run_turn(
...
+                        None,
+                        true,
+                    );
+                }
+            }
+
+            // Reserve an id atomically only for normal RTDL. Concurrent sessions
+            // cannot observe or dispatch the same id. A failed expansion may
+            // leave a harmless gap, but an id is never reused.
+            let plan_id = (plan_seq.fetch_add(1, Ordering::Relaxed) + 1).to_string();
```
enkerewpo, line 1411: **[P1] Plan IDs collide after a Pilot restart**

These IDs are generated only by an in-process counter, so a restarted Pilot starts again at `1` while Executor may still hold active plans created by the previous Pilot process or by another client. Submitting the same ID causes `PlanRuntime::register_plan` to replace the existing `PlanRun`, leaving two real executions with shared or lost cancellation state. Please use IDs that are unique across process lifetimes, or make Executor atomically reject duplicate active IDs.

system/executor/src/service.rs
```diff
@@ -461,6 +477,94 @@ impl RobonixSystemExecutorCancelAllPlans for ExecutorServiceImpl {
...
+            success: true,
+            plans_json: self.runtime.active_plans_json().await,
+            error: String::new(),
+        }))
+    }
+}
+
+#[tonic::async_trait]
+impl RobonixSystemExecutorControlPlan for ExecutorServiceImpl {
+    async fn control_plan(
```
enkerewpo, line 503: **[P2] `Execute` and plan control have a registration race**

`Execute` registers the plan inside its spawned task, after the RPC is already able to return its stream. A client can therefore issue `ListActivePlans` or `ControlPlan` immediately after `Execute` returns and observe the plan as absent; `cancel` then reports a successful already-finished no-op even though the background task subsequently registers and runs the plan. Please register the plan before returning the Execute response, or introduce a durable registration boundary that control requests wait for.

system/pilot/src/service.rs
```diff
@@ -291,32 +499,81 @@ impl RobonixSystemPilot for PilotServiceImpl {
...
+    .map_err(|error| format!("stop could not reach Executor: {error:#}"))?;
+    let mut client = RobonixSystemExecutorControlPlanClient::new(channel);
+    client
+        .control_plan(ControlPlanRequest {
+            action: "cancel_all".into(),
+            wait_ms: 5_000,
+            ..Default::default()
+        })
+        .await
+        .map(|response| response.into_inner().success)
```
enkerewpo, line 563: **[P2] `success` is being treated as `completed`**

`ControlPlanResponse` distinguishes request acceptance (`success`) from actual termination (`completed`), but this helper returns only `success`. When Executor times out waiting for cancellation it returns `success=true, completed=false`, and the caller still reports `stop completed`. Please inspect `completed` and preserve a timeout/still-stopping result instead of claiming completion.

enkerewpo (commented):
Four cancellation and lifecycle correctness findings.

enkerewpo:
Full Webots verification is now green on the unchanged PR head `9e676ac4629a40ab1692958f915dff00632dbec4`:

- Executor control-plane isolation
- full scenario suite
- Scene spatial/semantic map save and load persistence
- boot/shutdown lifecycle validation
- scenario rerun after lifecycle restart
- report generation and report-site publication

Run: https://github.com/syswonder/robonix/actions/runs/29720993246

The previous red manual run was caused by Mapping treating optimized occupancy pixels as map identity. That fix was reviewed through `syswonder/service-map-rbnx#11` and merged separately; no Mapping code or branch rename was added to this PR.

enkerewpo:
since we are refactoring, this PR will be closed now and re-issue later

PR #111, opened 2026-06-28
feat(vitals): Soma health stream ingestion, component-tree normalization, selector    threshold matching, and state-change push

AuYang261:
## Summary

  `dev-vitals` → `dev`. Adds the Vitals service — the robot health monitoring subsystem that consumes Soma's unified `SomaHealthSnapshot` gRPC stream, normalizes it through a component-tree topology into per-body `BodyHealth`
  entries, applies selector-based threshold rules, and pushes `VitalsSnapshot`  to Pilot / Sentinel / Liaison only on state transitions.

  ## Changes

  ### Soma ingestion + normalization (`soma_ingest.rs`)
  - Consumes `SomaHealthSnapshot` via gRPC `StreamHealth` server_stream.
  - Groups `ComponentStatus[]` tree by root's direct children into independent  `BodyHealth` entries — one arm, one chassis, one battery, etc. No hardcoded robot topology on the Vitals side.
  - Maps `ActuatorState` → `BodyComponent`, `PowerSourceState` → `PowerState`, `SafetyState.aggregate_state` → `BodyHealth.state`, `FaultState` → per-joint fault labels, `Metric[]` → `ComponentHealth` (when a threshold rule matches).

  ### Selector threshold matching
  - Three-level priority: exact `component_id` (3) > `component_id_glob` (2) > `kind` (1). Configured via `--thresholds-path` YAML (Soma selector format).
  - Falls back to built-in defaults covering JOINT / BATTERY / SENSOR / WHEEL when no rules file is provided.
  - `vendor_error_code` passed through untranslated; readable fault labels appended to `BodyHealth.message`.

  ### Mock Soma + Piper hardware bridge (`mock_soma.rs`, `piper_bridge.py`)
  - `--mock-soma` flag with deterministic scenarios: `ramp` (joint_1 temperature ramp → WARN → ERROR → recovery), `fault`, `toggle`, `mixed`.
  - `--mock-soma-interval-ms` configures tick rate.
  - Optional `--mock-soma-piper-can` spawns `piper_bridge.py` as a Python subprocess that reads real Piper arm data via `piper_sdk` CAN interface and merges it into the mock `SomaHealthSnapshot`.

  ### gRPC service surface (`service.rs`)
  - `GetVitals` (rpc): returns latest cached `VitalsSnapshot`.
[...]

PR #141, opened 2026-07-12
feat(atlas): make provider namespaces advisory

enkerewpo:
## Problem

Atlas treated a provider namespace as an authorization boundary and rejected every contract outside that prefix. This prevents one provider from implementing shared contracts: for example, a camera provider could not expose a standard health contract without creating a separate shadow provider.

The provider is already identified by `provider_id`; multiple providers are expected to implement the same `contract_id`.

## Design

- Keep `namespace` as the provider's primary grouping for discovery and diagnostics.
- Add `cross_namespace = true` to contract metadata for intentionally shared contracts.
- Accept other namespace mismatches, log a warning, and expose `Capability.namespace_mismatch` instead of blocking startup.
- Emit the same advisory warning from robonix-api and show it in `rbnx caps -v`, `describe`, and JSON output.

## Compatibility and migration

This is an additive wire change. Existing providers and per-namespace `*/driver` TOMLs continue to work unchanged.

After merge, Vitals #111 can mark its shared health contract as cross-namespace and let the existing hardware provider expose health directly. A later PR can add one built-in lifecycle contract and one built-in optional health contract; legacy package-local driver/health contracts must remain supported with migration notices.

## Validation

- GitHub CI: Rust, English-only docs, security preflight, and Webots report checks passed.
- macOS: workspace formatting, clippy, and all-target tests passed.
- Jetson AGX Orin (aarch64): Atlas and `rbnx` clippy/tests passed; release binaries were installed.
- robonix-api: sdist and wheel were rebuilt from the updated proto, installed, and verified to expose both additive fields.
- Isolated Atlas process on `127.0.0.1:54051` with the installed Python API:
  - two camera providers exposed the same legacy camera driver contract;
  - a shared health contract produced no namespace diagnostic;
  - an unmarked cross-namespace contract remained callable and produced exactly one diagnostic.

No robot stack or motion component was started during this validation.

## Related

- Vitals: #111
- Documentation: syswonder/robonix-book#8

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,8 +234,11 @@ Dive deeper:
 ## Ecosystem
 
 Robonix is built from small, swappable **packages**, each implementing one or
-more capability contracts under a `robonix/<kind>/<area>/*` namespace. The
-contract definitions are documented in the
+more capability contracts normally grouped under a primary
+`robonix/<kind>/<area>/*` namespace. Shared framework contracts may be
+implemented across those provider namespaces. Namespace mismatches are
+diagnostic rather than a runtime authorization boundary. The contract
+definitions are documented in the
 [interface catalog](https://github.com/syswonder/robonix-book/blob/main/src/interface-catalog/index.md).
 
 This repository contains the core runtime, built-in reference services, and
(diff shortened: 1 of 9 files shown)


enkerewpo (commented):
OK to merge

PR #142, opened 2026-07-13
feat: standardize core arm primitive contracts

enkerewpo:
## Problem

Arm providers currently define common lifecycle, joint-state, command, and end-pose contracts inside individual packages. This makes basic arm state discovery vendor-specific and prevents Soma from consuming a stable embodiment interface.

## Change

Add standard core contracts for:

- arm driver lifecycle
- joint-state output
- joint command input
- end-pose output

Vendor-specific status and command messages remain provider extensions. The end-pose contract intentionally preserves the deployed Piper v1 `geometry_msgs/Pose` wire type.

## Impact

Arm providers can implement shared Robonix contracts while remaining independently identifiable by provider ID. Existing package-local declarations remain compatible.

## Validation

- Proto, MCP, ROS 2, and documentation code generation passed.
- `cargo test -q -p robonix-codegen` passed.
- Merged code generation with the deployed Piper package-local capabilities passed without duplicate-contract conflicts.

## Discussion

Soma runtime aggregation of joint states is intentionally excluded and will be proposed in a separate draft PR.

### DIFF
--- a/capabilities/primitive/arm/driver.v1.toml
+++ b/capabilities/primitive/arm/driver.v1.toml
@@ -1,0 +1,10 @@
+# Lifecycle interface every arm primitive implementation must declare.
+[contract]
+id = "robonix/primitive/arm/driver"
+version = "1"
+kind = "primitive"
+idl = "lifecycle/srv/Driver.srv"
+description = "Lifecycle control interface for arm primitive providers."
+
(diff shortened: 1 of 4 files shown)


lhw2002426 (approved):
I think this is reasonable

PR #144, opened 2026-07-13
feat(webots): optimize GPU browser streaming

kaileliu:
- Improve GPU-backed Webots streaming while preserving the selected `ROBONIX_GPU_ID`.
- Add a cached browser viewer to prevent external asset loading stalls.
- Add an optimized WebSocket proxy for lower-latency real-time robot visualization.
- Add a keepalive process to keep the Webots simulation broadcasting live updates.
- Document remote browser access through SSH port forwarding.


enkerewpo:
@robonix-ci test

enkerewpo:
Thanks for the improvement. I see three issues to address before merging:

1. Why is the keepalive needed? Webots already starts in `realtime` mode and the viewer uses `broadcast`. Sending `real-time:1000` every 200 ms overrides legitimate pause/step operations. Please remove it, make it opt-in/state-aware, or document a reproducible case that requires it.
2. Custom ports do not work consistently: the viewer HTML still defaults to `1235`, and the viewer server always binds to `8080` while port publishing is ignored with the default host network.
3. The entrypoint does not verify that the viewer, proxy, and keepalive processes remain alive, so `start.sh` can report the stream as ready after a helper has already failed.

Please also add a small streaming smoke test covering custom ports and helper readiness; the current CI scenarios do not exercise this path.

PR #149, opened 2026-07-15
feat(rbnx): improve init/package-new templates and validate skill CAP…

lhw2002426:
…ABILITY.md

- init/package-new: emit YAML-safe manifests (quote description with colons) and align skill provider/IDL lib directory naming (test_skill_skill, capabilities/lib/<name>).
- package-new: generate CAPABILITY.md for skill packages.
- package-new: clarify --path semantics (parent dir for <name>/).
- validate: require CAPABILITY.md for skill packages.

### DIFF
--- a/tools/rbnx/src/cmd/init.rs
+++ b/tools/rbnx/src/cmd/init.rs
@@ -21,6 +21,102 @@ fn robot_catalog_name(name: &str) -> String {
     format!("robonix.robot.{}", suffix.replace('-', "."))
 }
 
+fn robot_slug(name: &str) -> String {
+    name.replace('-', "_")
+}
+
+fn deployment_manifest(name: &str, catalog_name: &str) -> String {
+    format!(
+        r#"manifestVersion: 1
+name: {name}
+
+catalog:
+  name: {catalog_name}
+  version: 0.1.0
+  description: "TODO: describe this robot deployment."
+  license: Apache-2.0
+  tags:
+    - robot
+    - deploy
+    - robonix
+  maintainers:
+    - Your Name <you@example.com>
+
+env: {{}}
+
+system:
+  atlas:
+    listen: 127.0.0.1:50051
+    log: info
+  executor:
+    listen: 127.0.0.1:50061
+    log: info
+  soma:
+    robot_yaml: ./soma.yaml
+  vitals:
+    listen: 127.0.0.1:50093
+  pilot:
+    listen: 127.0.0.1:50071
+    log: info
+    vlm:
+      upstream: ${{VLM_BASE_URL}}
+      api_key: ${{VLM_API_KEY}}
+      model: ${{VLM_MODEL}}
+  liaison:
+    listen: 127.0.0.1:50081
+    log: info
+
+primitive: []
+
+service: []
+
+skill: []
+"#
+    )
+}
+
+fn soma_manifest(name: &str, robot_file: &str, robot_id: &str) -> String {
+    format!(
+        r#"urdf:
+  path: ./urdf/{robot_file}
+  root_link: base_link
+  model_name: {robot_id}
+
+robot:
+  id: {robot_id}
+  display_name: "{name}"
+  family: mobile_robot
+  root_part: base
+  dimensions: {{ length_m: 0.50, width_m: 0.50, height_m: 1.00 }}
+  mass_kg: 20.0
+  exports: []
+  components:
+    - id: base
+      type: mobile_base
+      urdf_link: base_link
+      exports: []
+
+description:
+  summary: "TODO: describe this robot deployment."
+  can_do: []
+  cannot_do: []
+  notes:
+    - "Fill in robot.exports and robot.components as you add primitives, services, and skills."
+    - "Keep provider_id values in soma.yaml aligned with the corresponding name entries in robonix_manifest.yaml."
+"#
+    )
+}
+
+fn minimal_urdf(robot_id: &str) -> String {
+    format!(
+        r#"<?xml version="1.0"?>
+<robot name="{robot_id}">
+  <link name="base_link"/>
+</robot>
+"#
+    )
+}
+
 pub async fn execute(name: &str, path: Option<&Path>) -> Result<()> {
     validate_name(name)?;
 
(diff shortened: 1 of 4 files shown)


enkerewpo:
> …ABILITY.md
> 
> * init/package-new: emit YAML-safe manifests (quote description with colons) and align skill provider/IDL lib directory naming (test_skill_skill, capabilities/lib/).
> * package-new: generate CAPABILITY.md for skill packages.
> * package-new: clarify --path semantics (parent dir for /).
> * validate: require CAPABILITY.md for skill packages.

the PR description seems incomplete

tools/rbnx/src/cmd/package_new.rs
```diff
@@ -91,109 +161,228 @@ start: bash scripts/start.sh
...
+    let base_dir: PathBuf = if let Some(p) = path {
+        if p.is_absolute() {
+            p.to_path_buf()
+        } else {
+            std::env::current_dir()?.join(p)
+        }
+    } else {
+        std::env::current_dir()?.join(role_dir(pkg_type))
+    };
+    let pkg_dir = base_dir.join(name);
```
enkerewpo, line 303: **[P1] Preserve the existing `--path` contract.** Before this PR, `rbnx package-new my_cam --path ./primitives/my_cam` treated the argument as the exact package directory. This unconditional join now creates `./primitives/my_cam/my_cam`, breaking existing documentation and automation. Please keep exact-target paths working (for example, detect when the final component already equals `name`, optionally with a migration warning) while also supporting the new parent-directory form, and add regression coverage for both forms.

tools/rbnx/src/cmd/validate.rs
```diff
@@ -24,6 +24,18 @@ pub async fn execute(path: Option<PathBuf>) -> Result<()> {
     let detected = manifest::detect_and_load(&package_root, None)?;
     let summary = detected.manifest.validate_and_summarize()?;
 
+    let is_skill_package = summary.name.starts_with("robonix.skill.")
+        || summary
+            .capabilities
+            .iter()
+            .any(|cap| cap.starts_with("robonix/skill/"));
+    if is_skill_package && !package_root.join("CAPABILITY.md").is_file() {
+        anyhow::bail!(
```
enkerewpo, line 33: **[P1] Do not invalidate every existing skill package during this migration.** Skills created before `CAPABILITY.md` was introduced can have valid manifests and executable capabilities but will now fail `rbnx validate` immediately. This needs an incremental compatibility path: emit a clear deprecation warning for legacy packages, and only make the document mandatory behind a new manifest/schema version (or after an explicit migration).

tools/rbnx/src/cmd/init.rs
```diff
@@ -64,37 +51,118 @@ system:
...
+  id: {robot_id}
+  display_name: "{name}"
+  family: mobile_robot
+  root_part: base
+  dimensions: {{ length_m: 0.50, width_m: 0.50, height_m: 1.00 }}
+  mass_kg: 20.0
+  exports: []
+  components:
+    - id: base
+      type: mobile_base
```
enkerewpo, line 95: **[P1] A generic `rbnx init` must not invent a mobile-base embodiment.** These generated values assert `mobile_robot`, a `mobile_base`, dimensions, and mass even for arm-only, quadruped, or other deployments. Soma and Pilot may then treat fabricated embodiment facts as authoritative. Please either collect the embodiment type explicitly or generate a neutral, unmistakably incomplete template that requires the integrator to define the real component tree before use.

tools/rbnx/src/cmd/package_new.rs
```diff
@@ -205,10 +394,16 @@ if __name__ == "__main__":
         pkg_dir.display()
     ));
     output::sub_step("package_manifest.yaml");
+    output::sub_step("config.spec");
     output::sub_step("scripts/build.sh");
     output::sub_step("scripts/start.sh");
-    output::sub_step(&format!("{module_name}/main.py  (provider skeleton)"));
-    output::sub_step("capabilities/  (.gitkeep)");
+    output::sub_step(&format!("capabilities/driver.v1.toml"));
```
enkerewpo, line 400: **[P1] Fix the Rust CI failure.** Clippy rejects this as `useless_format`; the argument is already a string literal. Use `output::sub_step("capabilities/driver.v1.toml");` so the required `-D warnings` check passes.

enkerewpo (changes requested):
The template direction is useful, but this revision introduces two backward-incompatible migrations, generates embodiment-specific Soma data from a generic command, and currently fails the Rust check. Please address the inline findings before merging.

tools/rbnx/src/cmd/package_new.rs
```diff
@@ -32,57 +33,126 @@ fn python_module_name(name: &str) -> String {
...
+    )
+}
+
+fn driver_contract(ns_kind: &str, name: &str) -> String {
+    format!(
+        r#"[contract]
+id      = "robonix/{ns_kind}/{name}/driver"
+version = "1"
+kind    = "{ns_kind}"
+idl     = "lifecycle/srv/Driver.srv"
```
enkerewpo, line 110: **[P1] Keep the generated package structure aligned with the lifecycle spec and docs.** The current `dev-next` `robonix-api` documentation describes package-local `*/driver` TOMLs as a backward-compatible legacy form and says new packages will move to a built-in shared lifecycle contract. This scaffold instead creates a new package-local driver contract for every package, making the compatibility form canonical again. Either land/use the built-in lifecycle contract before removing this file, or explicitly keep this structure as the current canonical format and update `robonix-book` in the same change. The CLI, runtime contract registry, template repository, and developer guide must describe one package shape.

enkerewpo (commented):
I also compared this PR with the existing `dev-next` scaffold and the current `robonix-book` developer guide. The implementation can be reused, but it must be reconciled rather than merged as a second, divergent package format: retain the `dev-next` manifest metadata tests and update the package-structure/CLI documentation together with the behavior.

enkerewpo (commented):
Compatibility boundary for this PR:

- `config.spec` and `CAPABILITY.md` are recommended documentation artifacts. New scaffolds should generate them, and validation may warn when they are missing, but their absence must not block validation, build, boot, registration, or execution of existing packages.
- Runtime-required manifest and contract data remains mandatory where the package actually depends on it: a valid package manifest, package identity/entrypoints required by that package type, and resolvable contract/IDL definitions for declared private capabilities.
- Packages implementing only built-in standard contracts must not be forced to copy private contract files.
- Existing `--path` behavior, legacy manifest fields such as `vendor`, existing skill layouts, and package-local lifecycle contracts must remain supported during migration. Deprecations should be warnings with an explicit replacement path, not hard failures.

Please make the validator, generated templates, tests, and `robonix-book` describe this same required-versus-recommended split.

PR #150, opened 2026-07-16
feat(scene): epoch-safe semantic map persistence keyed by save tokens

HeartLinked:
## Problem

Semantic objects persisted by Scene were not epoch-aware. Rows written across different map-frame epochs—reset, re-mapping, or reboot—accumulated under the same partition.

As a result, `Load` could restore objects anchored to a frame that no longer existed, causing the “objects drift outside the map” family of bugs: H1, H2, H3, M1, and M2 from the #137 review, following the agreed split.

## Mechanism

* **Save-token sidecar**

  * Every `Save` writes the object snapshot into a fresh Milvus partition:

    ```text
    <map_id>__s<seq>
    ```
  * A sidecar JSON records the partition containing the map’s semantic snapshot.
  * The sidecar is committed only after the write is verified as complete.
  * The previous snapshot is purged after the sidecar commit succeeds.

* **Strict Load behavior**

  * `Load` restores exactly the partition referenced by the sidecar.
  * Maps containing only legacy data have no sidecar, so `Load` restores no semantic objects and reports this explicitly in the response.

* **Runtime epochs**

  * The lifecycle watcher now acts on generation bumps.
  * It flushes derived objects and marks rooms stale.
  * Derived data remains expendable; user assets remain preserved.
  * Watcher actions and facade operations are serialized under a shared `ops_lock`.

* **Generation independence**

  * Epoch safety does not depend on Mapping’s generation broadcast.
  * The broadcast supports the in-session response only.
  * Related change: `service-map-rbnx#3`, merged.

## Commits

1. `feat(scene)`: Snapshot storage layer

   * `map_meta` sidecar
   * Explicit partitions
   * Strict load behavior

2. `feat(scene)`: Facade `Save` / `Load` / `Delete`

   * Commit ordering
   * `409` guards
   * Reserved namespaces
   * Self-robot exclusion

3. `feat(scene)`: Runtime epoch response

   * Watcher actor
   * `.live` partition
   * Boot sweep
   * Builder gating

4. `test(scene)`: 23 new tests

5. `docs(scene)`: README updates

### Container Suite

* **106 passed**
* **1 skipped**
* Baseline: 83 tests
* New: 23 tests

### Webots End-to-End Testing

Validated across 14 boots:

* Full-chain `Save`
* Mapping-mode re-save returns `409`
* Cold-boot `Load` ×3
* No-sidecar `Load` with legacy semantics
* Port `8091` reset triggers:

  * Watcher flush
  * Room stale marking
  * Exactly one epoch bump
* UI `Load` from the `/user` page

### Known Gap

The `load → add room → re-save` path was blocked by an upstream RTAB-Map second-save issue, tracked separately. The behavior is covered by unit tests.

Follows up on #137 and addresses the epoch-related issue family identified in its review, following the agreed split.

### DIFF
--- a/system/scene/scene_service/map_meta.py
+++ b/system/scene/scene_service/map_meta.py
@@ -1,0 +1,139 @@
+# SPDX-License-Identifier: MulanPSL-2.0
+"""Scene-owned per-map semantic snapshot metadata (the epoch sidecar).
+
+mapping owns a saved map's spatial artifact; scene owns the matching semantic
+snapshot (objects + room annotations). The sidecar records, per `map_id`,
+WHICH object partition holds the snapshot written by the Save that produced
+the artifact on disk. Every Save allocates a fresh partition token
+(`"<map_id>__s<seq>"`) and repoints the sidecar at it, so:
+
+- Load restores exactly the rows written together with the loaded artifact —
+  never rows from an earlier build of a same-named map, whose map frame no
+  longer exists (the mis-anchored / off-map restore bug).
+- A map with NO sidecar (saved before this mechanism, or a foreign DB)
+  restores nothing rather than coordinates in a dead frame.
+
+One JSON file per map under `base_dir`, written atomically (tmp + replace).
+`mapping_generation` / `mapping_mode` are recorded for diagnostics when the
+lifecycle broadcast is available at save time; the mechanism itself never
+depends on them.
+"""
+from __future__ import annotations
+
+import json
+import logging
+import os
+import threading
+import time
+from dataclasses import asdict, dataclass
+from pathlib import Path
+from typing import Optional
+
+from .map_binding import sanitize_map_id
+
+log = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class MapSemanticMeta:
+    """Sidecar record for one saved map's semantic snapshot."""
+    map_id: str
+    object_partition: str        # milvus partition holding the snapshot rows
+    save_seq: int                # monotonic per map; allocates the partition
+    saved_at_unix: float
+    mapping_generation: Optional[int] = None  # broadcast gen at save (diagnostic)
+    mapping_mode: str = ""                    # broadcast mode at save (diagnostic)
+
+    def to_json(self) -> dict:
+        return asdict(self)
+
+    @classmethod
+    def from_json(cls, d: dict) -> "MapSemanticMeta":
+        known = set(cls.__dataclass_fields__)
+        return cls(**{k: v for k, v in d.items() if k in known})
+
+
+class MapMetaStore:
+    """Per-map sidecar files: `<base_dir>/<map_id>.json`.
+
+    All methods are synchronous; the internal lock guards individual file
+    ops only — the `next_partition` → persist rows → `write` sequence is NOT
+    atomic here, and callers serialize whole Save/Load/Delete operations
+    (the web facade's `ops_lock`). A file that cannot be parsed is treated
+    as absent with a WARNING: restoring nothing is the safe outcome, and the
+    bytes are left in place for inspection."""
+
+    def __init__(self, base_dir: str) -> None:
+        self._lock = threading.Lock()
+        self._base = Path(base_dir).expanduser()
+        self._base.mkdir(parents=True, exist_ok=True)
+
+    def _path(self, map_id: str) -> Path:
+        return self._base / f"{sanitize_map_id(map_id)}.json"
+
+    def read(self, map_id: str) -> Optional[MapSemanticMeta]:
+        """The sidecar for `map_id`, or None when absent/unreadable. A
+        malformed record must never crash a Load — it degrades to
+        "no snapshot", which restores nothing."""
+        path = self._path(map_id)
+        with self._lock:
+            if not path.exists():
+                return None
+            try:
+                data = json.loads(path.read_text(encoding="utf-8"))
+                meta = MapSemanticMeta.from_json(data)
+                if not meta.object_partition:
+                    raise ValueError("empty object_partition")
+                return meta
+            except Exception as e:  # noqa: BLE001
+                log.warning(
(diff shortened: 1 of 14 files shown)


enkerewpo:
@robonix-ci test

system/scene/scene_service/web.py
```diff
@@ -789,16 +893,56 @@ async def maps_save(request) -> JSONResponse:
...
+                    for part in stale_parts:
+                        try:
+                            await asyncio.to_thread(object_store.delete_map, part)
+                        except Exception as e:  # noqa: BLE001
+                            # Orphan rows cost disk, not correctness — the
+                            # sidecar no longer points at them.
+                            out["stale_partition_cleanup_error"] = str(e)
+                except Exception as e:  # noqa: BLE001
+                    object_persist_error = str(e)
+                    out["object_persist_error"] = object_persist_error
```
enkerewpo, line 938: [P1] Please do not leave this operation as `ok=true` after the semantic snapshot commit fails. The Mapping RPC may already have saved geometry, but the public Save transaction has not saved geometry + rooms + objects as one unit; returning HTTP 200 lets the UI/operator treat an incomplete artifact as complete. Return an explicit failed/partial state (and test it) so callers must retry before continuing.

system/scene/scene_service/web.py
```diff
@@ -893,13 +1042,69 @@ async def maps_load(request) -> JSONResponse:
...
             try:
-                out["objects_restored"] = await _restore_scene_objects(map_id)
+                async with registry.lock():
+                    flushed = registry.clear_objects()
+                if flushed:
+                    out["objects_flushed"] = flushed
             except Exception as e:  # noqa: BLE001
-                out["object_restore_error"] = str(e)
-            _set_map_binding(map_id, "localization", "ui_load")
+                out["object_flush_error"] = str(e)
```
enkerewpo, line 1056: [P1] A registry flush failure must abort this Load before annotations or objects are rebound. Continuing here restores the new map snapshot into a registry that may still contain objects anchored to the previous frame, which recreates the mixed-epoch state this PR is meant to prevent. Please fail the load and add a `clear_objects()` failure regression test.

system/scene/scene_service/web.py
```diff
@@ -893,13 +1042,69 @@ async def maps_load(request) -> JSONResponse:
...
+                    # genuinely empty one, in the operator-visible detail —
+                    # saving on top of an unnoticed failed restore would
+                    # commit an empty snapshot.
+                    out["object_restore_error"] = str(e)
+                    out["detail"] = (
+                        f"{out.get('detail') or 'loaded'}; RESTORE FAILED: "
+                        f"{e} — saved objects were NOT loaded; retry the "
+                        "load before saving this map"
+                    )
+            _set_map_binding(map_id, "localization", "ui_load", generation=gen)
```
enkerewpo, line 1107: [P1] Do not commit the new Scene binding when object restore failed above. `out.ok` remains true, the registry may be empty, and the next Save is then allowed to publish an empty fresh partition and move the sidecar away from the last valid snapshot. Keep the previous binding (or mark this binding restore-incomplete), return failure, and cover the load-failure-then-save sequence.

system/scene/scene_service/service.py
```diff
@@ -972,49 +1008,84 @@ async def _lifecycle_watch(
...
+                        # Losing the stale marker is recoverable (next
+                        # boot's load-time epoch check re-judges); losing
+                        # this watcher would silence ALL drift handling.
+                        log.error(
+                            "[scene-anno] stale marking failed (annotations "
+                            "may show as fresh until restart): %s", e,
+                        )
+                # Advance the live epoch: the response fired once for this
+                # bump and the next bump is measured against it.
+                live_binding["generation"] = live_gen
```
enkerewpo, line 1052: [P1] `_flush_registry()` swallows its exception, but this line still acknowledges the new generation. The next tick therefore sees no mismatch and never retries, leaving mis-anchored objects indefinitely. Make the flush result observable and advance `live_binding["generation"]` only after it succeeds (or retain a pending generation for retry); add a failing-flush watcher test.

system/scene/scene_service/map_meta.py
```diff
@@ -0,0 +1,139 @@
...
+        malformed record must never crash a Load — it degrades to
+        "no snapshot", which restores nothing."""
+        path = self._path(map_id)
+        with self._lock:
+            if not path.exists():
+                return None
+            try:
+                data = json.loads(path.read_text(encoding="utf-8"))
+                meta = MapSemanticMeta.from_json(data)
+                if not meta.object_partition:
```
enkerewpo, line 85: [P1] Please validate the sidecar identity, not only that `object_partition` is non-empty. As written, `read("map_a")` accepts parseable JSON containing `map_id="map_b"` and `object_partition="map_b__s1"`, so Load can restore another map's objects into `map_a`. Require the sanitized stored map id to equal the requested id and the partition/save sequence to match that map's token format; treat mismatches as corrupt and add a regression test.

enkerewpo (changes requested):
The save-token model is the right direction, but the current error paths can still acknowledge an incomplete semantic transaction or permanently stop retrying an epoch flush. These cases can reintroduce mixed-frame objects or overwrite the last valid object snapshot, so they need to be closed before merge.

enkerewpo:
Robonix will undergo a repository history cleanup. To keep this work available while avoiding further development on fork history that will soon be rewritten, we will preserve the current PR head exactly in the main repository as `feat/scene-epoch-persistence` at `3d5cf54a868f40e8d5bebbbfe47b367e47d6a048`.

This PR will then be closed without merging it into `dev`. Closing it is an administrative migration, not an acceptance or rejection of the implementation. The existing commits and authorship will remain intact on the preserved branch.

After the history cleanup, the contributor will be added to `syswonder/robonix` and can continue directly on `feat/scene-epoch-persistence`. Once the work is ready and reviewed, please open a new PR from that branch to `dev`.

PR #153, opened 2026-07-16
fix(scene): honor ROS contracts and cache model builds

Origamii520:
## Summary

- discover the canonical lidar3d and chassis odometry capability contracts
- preserve Atlas-declared ROS QoS for sensor, odometry, and latched map topics
- install and forward CycloneDDS settings into the Scene container
- forward provider bind/advertise settings and the persistent annotations path
- use the Docker daemon's bundled frontend by default instead of a registry-resolved syntax image
- prefetch and cache the exact OpenCLIP checkpoint on the host, then use its local path in Docker and native runtimes
- keep mirror selection optional with a canonical Hugging Face fallback

## Motivation

The Unitree Go2 deployment uses an interface-bound CycloneDDS graph and the canonical Robonix primitive contracts. Without these changes, Scene can miss the robot's lidar/odometry streams or subscribe with incompatible QoS.

The previous image build also resolved both its Dockerfile frontend and OpenCLIP model metadata through external registries during BuildKit execution. In restricted or proxied environments that made otherwise cached builds fail. Scene now stages the exact model file through the existing host retry/cache path and avoids the frontend lookup unless an explicit override is provided.

## Validation

- Scene ROS/build contract suite: 10 tests passed
- shell syntax and committed diff whitespace checks passed
- full `robonix-scene:latest` image built successfully, including local OpenCLIP validation and generated `map` interface build
- no Scene service, ROS graph, robot hardware, or motion command was started

Related deployment: https://github.com/syswonder/robot-unitree-go2


enkerewpo:
@robonix-ci test

enkerewpo:
LGTM

enkerewpo (approved):

enkerewpo:
@all-contributors please add @Origamii520 for code

PR #165, opened 2026-07-20
fix(speech): make playback and provider selection reliable

enkerewpo:
## What changed

Backport the remaining isolated Speech runtime fixes from `dev-next` to `dev`:

- serialize concurrent playback requests per speaker provider while allowing different speakers to run independently
- use Tencent's basic `16k_zh` realtime ASR engine by default
- return only gRPC speaker providers that can actually accept playback requests

The revised-ASR transcript deduplication change was audited but is already present in `dev`, so it is intentionally not duplicated here. This PR does not include the Sherpa installation change from #164.

## Why

Concurrent TTS calls could overlap on one physical speaker, the previous Tencent engine default could select a product without the expected free quota, and speaker discovery could report providers that were not playable through the Speech service.

## Validation

- Speech unit tests: 5 passed
  - per-speaker playback serialization
  - speaker discovery filtering
  - Tencent ASR URL/signature behavior
- `python3 -m compileall -q services/speech/speech_service services/speech/tests`
- `git diff --check`

All commits use `wheatfox <wheatfox17@icloud.com>`.

### DIFF
--- a/services/speech/README.md
+++ b/services/speech/README.md
@@ -29,7 +29,7 @@ service:
     config:
       speech_backend: tencent
       tencent_asr_appid: "1234567890"
-      tencent_asr_engine: 16k_zh_en
+      tencent_asr_engine: 16k_zh
       tencent_tts_voice_type: 502003
       tencent_tts_region: ap-guangzhou
 ```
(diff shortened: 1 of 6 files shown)


enkerewpo (commented):
LGTM

PR #169, opened 2026-07-21
ci: harden dev Webots runner isolation

enkerewpo:
## Summary

Harden the `dev` Webots pipeline against the runner failures seen across recent push and ChatOps runs.

## Root causes

- A valid RTAB-Map save/load cycle could differ by one boundary cell, while the verifier required byte-for-byte grid dimensions.
- A rerun reused the same runtime directory and failed to remove root-owned Scene and Mapping artifacts from the previous attempt.
- Earlier runs started GPU-dependent build work while an unrelated process had consumed nearly all GPU memory.
- The offline scenario simulator overwrote repeated regex checks with a literal placeholder, causing `@robonix-ci test` to fail before Webots.

## Changes

- Serialize heavy Webots jobs per selected GPU and wait for a configurable minimum free-memory threshold.
- Keep Voiceprint build-time model verification on CPU so Webots and Scene retain the selected GPU.
- Include `github.run_attempt` in runtime, container, Compose, map, and Robonix-home identities.
- Remove container-owned runtime data after artifacts are uploaded.
- Allow a bounded two-cell occupancy-grid boundary quantization difference while retaining the existing resolution, origin, database, and non-empty-map checks.
- Synthesize and accumulate regex-matching fixture text in the offline scenario harness.

## Local validation

- Current `dev` scenario simulation: 15/15
- PR #121 scenario simulation: 17/17
- Python bytecode compilation for both modified verifiers
- `actionlint` (with only the repository-specific self-hosted label acknowledged)
- Bash syntax validation for modified workflow scripts
- `git diff --check`
- Robonix commit authorship check

Full self-hosted Webots and ChatOps validation will be attached to this PR before it is marked ready.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -302,6 +302,9 @@ jobs:
   gpu_testing:
     name: Webots integration suite
     needs: security-gate
+    concurrency:
+      group: robonix-webots-gpu-${{ inputs.gpu_id || vars.ROBONIX_CI_GPU_ID || '0' }}
+      cancel-in-progress: false
     outputs:
       result: ${{ steps.scenarios.outputs.result }}
     # Optional per-run manual approval: configure the `gpu-testing` Environment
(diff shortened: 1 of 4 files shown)


enkerewpo:
@robonix-ci test

enkerewpo:
## Validation update

The final head `90e4f86` has now passed all targeted validation:

- Full Webots workflow: https://github.com/syswonder/robonix/actions/runs/29797303078 (attempt 1, all jobs passed)
- Same run rerun with `github.run_attempt=2`: https://github.com/syswonder/robonix/actions/runs/29797303078/attempts/2 (all jobs passed)
- Post-run runner audit: no attempt directories, containers, or active Worker remained
- Real `@robonix-ci test` entry point: https://github.com/syswonder/robonix/actions/runs/29798887943 (authorization, PR merge, Webots, report, and status comment all passed)

The attempt-2 run directly covers the original root-owned runtime-directory collision instead of relying on a fresh run ID.

enkerewpo:
@robonix-ci test

PR #174, opened 2026-07-22
fix(runtime): make planning state-aware and RTDL control reliable

enkerewpo:
> [!NOTE]
> This PR recreates #140 after the repository history cleanup. The preserved implementation now lives on the main-repository branch `feat/pilot-harness-lifecycle`. Closing #140 was an administrative migration, not an acceptance or rejection of the work.

## Problem

Pilot previously mixed task execution, cancellation, steering, and user-facing replies in one RTDL lifecycle. It also planned from stale local state instead of refreshing the authoritative Executor, Soma, and Scene snapshots. With concurrent or persistent plans this caused cancel loops, duplicate or delayed replies, incorrect running-task counts, and plans that ignored the robot's current motion, gripper, room, or nearby-object state.

## Design

- Add `ControlPlan` and `ListActivePlans` as an explicit Executor control plane.
- Keep cancel, cancel-all, and stop-boundary operations outside ordinary RTDL nodes.
- Target one plan without canceling independent plans or the control request itself.
- Terminate canceled Bash process groups instead of only dropping their streams.
- Refresh Pilot from Executor's authoritative active-plan snapshot before each planning round.
- Refresh embodiment state from Soma and environment state from Scene before planning.
- Add Scene's focused `get_robot_context` query so room, area, pose, and nearby objects do not depend on a truncated all-object listing.
- Scope steer replies to the newest input and emit one final reply at the actual terminal boundary.
- Preserve legacy Executor builtins for compatibility, while hiding them from the model.
- Bound completed-tree feedback and share one valid-image predicate across history and final-response serialization, including the 4096-byte final payload limit.

## Scope

This PR contains only Pilot/Executor planning-context, RTDL execution/control, steering, cancellation, history/image handling, and reply-lifecycle changes. Shared lifecycle Driver migration, Soma/Boot instance identity, Vitals, Liaison/audio, Atlas namespace policy, and general `rbnx boot` cleanup remain outside this experimental PR.

## Validation

Latest Pilot patch validation on the current PR head:

- `git diff --check`
- Pilot unit tests: 51 passed

Earlier integrated branch validation:

- Executor unit tests: 36 passed
- Pilot unit tests: 47 passed
- Soma unit tests: 27 passed
- Scene robot-context test: 1 passed after isolated `rbnx codegen --mcp --ros2`
- Total targeted tests: 111 passed, 0 failed

The branch also includes isolated Executor cancellation and natural-language RTDL integration scripts for full runtime validation. It intentionally remains an open PR so the experimental Pilot behavior can be reviewed and runtime-tested before entering `dev`.

## Migrated review context

- Original PR: #140
- The four original cancellation/lifecycle findings and the last full Webots result are restated in a comment below.
- Rebase/merge the latest `dev` and rerun the full interaction scenarios before final merge.


enkerewpo:
## Review context migrated from #140

The original PR had four cancellation and lifecycle correctness findings. The preserved branch has advanced since then, so each item must be re-evaluated against the current diff before it is considered resolved:

1. [A turn abort must not cancel plans owned by other sessions](https://github.com/syswonder/robonix/pull/140#discussion_r3575225812).
2. [Plan IDs must remain unique across Pilot restarts and concurrent clients](https://github.com/syswonder/robonix/pull/140#discussion_r3575225817).
3. [Execute and plan-control registration must not race](https://github.com/syswonder/robonix/pull/140#discussion_r3575225821).
4. [Control request acceptance must not be reported as completed termination](https://github.com/syswonder/robonix/pull/140#discussion_r3575225825).

The last full Webots run on the old PR head was green and covered Executor isolation, the full scenario suite, Mapping/Scene persistence, lifecycle restart, and report publication: [run 29720993246](https://github.com/syswonder/robonix/actions/runs/29720993246). That result is historical; this recreated PR now targets a newer `dev` and requires fresh validation after synchronization.

enkerewpo (commented):
Code review found one blocking isolation bug in the abort path: `abort_turn` was session-scoped at the Pilot API but called Executor `cancel_all`, so stopping one session could terminate independent RTDL plans owned by other sessions. Commit `288dd2bb` removes that global control-plane call and leaves cancellation to the active turn supervisor, which already targets only the forest owned by that session.

Validation after the fix: `cargo fmt --all -- --check`, `cargo test -p robonix-pilot` (51 passed), and `git diff --check`.

I also rechecked the per-turn broadcast ownership, targeted `cancel_plan`/`stop_at` flow, authoritative active-plan snapshot, bounded completed-tree feedback, and final payload/image handling. No additional blocking finding from this pass; full runtime cancellation/steer scenarios remain required before merge.

enkerewpo:
@robonix-ci test

PR #175, opened 2026-07-22
feat(keystone): add persistent user identity and access

enkerewpo:
## Summary

Add Keystone as a boot-managed Robonix system component and connect persistent user identity to Liaison, Voiceprint, Pilot task submission, hands-free voice sessions, `rbnx chat`, and Robonix Client.

## Why

Robonix previously had no durable account authority. A client-provided user hint could not prove which account was acting, voiceprint enrollment was not bound to a signed-in account, and administrative policy had no protected control plane.

## What changed

- define the canonical Keystone IDL and capability contracts in the shared Robonix capability tree;
- add a SQLite-backed Keystone service with WAL mode, foreign keys, Argon2 password hashes, opaque revocable sessions, roles, account enablement, and per-user voice guard policy;
- bootstrap the first administrator from an explicit environment password or a randomly generated mode-0600 one-time credentials file;
- preserve users, roles, sessions, policy, and voiceprint bindings across process restarts;
- add registration, login, logout, profile/password management, voiceprint binding and preview, and administrator RPCs;
- prevent deletion, disablement, or demotion of the final enabled administrator, including concurrent updates;
- register Keystone and its capabilities in Atlas, including the repository-level shared `robonix/lifecycle/driver`;
- launch Keystone as an `rbnx` built-in system component and include it in install, clean, deploy, and package-run paths;
- resolve authenticated text and voice turns to the canonical Keystone user before submitting to Pilot;
- enforce voice guard by matching the signed-in account's bound Voiceprint identity;
- require an authenticated session for hands-free control when Keystone is configured;
- extend Liaison voice-session and hands-free IDL with the opaque session token;
- keep deployments without a configured Keystone endpoint backward compatible;
- configure the Webots deployment for Atlas discovery by external clients.

## Security boundary

Clients discover Keystone through Atlas and connect to its generated capability endpoints. Session tokens are random opaque values; only their SHA-256 hashes are stored. Passwords use Argon2. A client-supplied user ID is never accepted as authentication.

This PR does not add Sentinel policy enforcement.

## Validation

- `cargo fmt --all -- --check` — passed;
- `cargo test -p robonix-keystone` — 9 tests passed;
- `cargo test -p robonix-liaison` — 15 tests passed;
- `cargo test -p robonix-cli` — 24 tests passed;
- `cargo clippy -p robonix-keystone -p robonix-liaison -p robonix-cli --all-targets -- -D warnings` — passed;
- Debian 13 WorkPC full Webots boot — 14 components active, including Keystone, Executor, Liaison, Scene, Speech, Voiceprint, Mapping, and Navigation;
- restart persistence — an existing account and session remained valid after a full-stack restart;
- macOS Client to WorkPC Robonix — connection, signup, login, profile, administrator authorization, role and voice-guard changes, authenticated text submission, audio routing, microphone capture, and speaker playback passed;
- stored voiceprint preview — the signed-in account retrieved its saved PCM sample and the client rendered and played it;
- authorization — a normal user was denied administrator RPCs.

## Dependencies

The matching Client integration is `syswonder/robonix-client#6`.

This PR targets `dev` and keeps the existing `dev-keystone` review branch.

### DIFF
--- a/system/keystone/src/config.rs
+++ b/system/keystone/src/config.rs
@@ -1,0 +1,150 @@
+// SPDX-License-Identifier: MulanPSL-2.0
+
+use std::path::PathBuf;
+
+use anyhow::{Context, Result};
+use clap::Parser;
+use serde::Deserialize;
+
+pub const DEFAULT_ATLAS_ENDPOINT: &str = "127.0.0.1:50051";
+pub const DEFAULT_LISTEN: &str = "0.0.0.0:50095";
+pub const DEFAULT_PROVIDER_ID: &str = "keystone";
+pub const KEYSTONE_NAMESPACE: &str = "robonix/system/keystone";
+
+#[derive(Debug, Clone)]
+pub struct KeystoneConfig {
+    pub atlas_endpoint: String,
+    pub listen: String,
+    pub provider_id: String,
+    pub database: PathBuf,
+    pub bootstrap_credentials_file: PathBuf,
+    pub bootstrap_admin_username: String,
+    pub bootstrap_admin_display_name: String,
+    pub bootstrap_admin_email: String,
+    pub bootstrap_admin_password: Option<String>,
+}
+
+#[derive(Parser, Debug)]
+#[command(name = "robonix-keystone", about = "Robonix user identity service")]
+pub struct Args {
+    #[arg(long, env = "ROBONIX_ATLAS_ENDPOINT")]
+    pub atlas: Option<String>,
+
+    #[arg(long, env = "ROBONIX_KEYSTONE_LISTEN")]
+    pub listen: Option<String>,
+
+    #[arg(long, env = "ROBONIX_KEYSTONE_PROVIDER_ID")]
+    pub provider_id: Option<String>,
+
+    #[arg(long, env = "ROBONIX_KEYSTONE_DATABASE")]
+    pub database: Option<PathBuf>,
+
+    #[arg(long, env = "ROBONIX_KEYSTONE_BOOTSTRAP_FILE")]
+    pub bootstrap_credentials_file: Option<PathBuf>,
+
+    #[arg(long, env = "ROBONIX_KEYSTONE_BOOTSTRAP_ADMIN_USERNAME")]
+    pub bootstrap_admin_username: Option<String>,
+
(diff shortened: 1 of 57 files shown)


enkerewpo:
@robonix-ci test

system/keystone/src/service.rs
```diff
@@ -0,0 +1,559 @@
...
+        let input = request.into_inner();
+        let target = self
+            .store
+            .list_users(&input.session_token)
+            .map_err(status)?
+            .into_iter()
+            .find(|user| user.user_id == input.target_user_id)
+            .ok_or_else(|| Status::not_found("target user does not exist"))?;
+        if target.voiceprint_enrolled {
+            self.delete_external_voiceprint(&target.user_id, "").await?;
```
kaileliu, line 268: Non-blocking follow-up: the external Voiceprint deletion happens before `store.admin_delete_user` enforces the final-enabled-admin invariant. Deleting the last admin therefore removes the external enrollment, then returns `permission_denied` while Keystone still records the user as enrolled. Please validate the local mutation before the external side effect, or add compensation that restores a consistent binding on local rejection.

system/keystone/src/config.rs
```diff
@@ -0,0 +1,150 @@
+// SPDX-License-Identifier: MulanPSL-2.0
+
+use std::path::PathBuf;
+
+use anyhow::{Context, Result};
+use clap::Parser;
+use serde::Deserialize;
+
+pub const DEFAULT_ATLAS_ENDPOINT: &str = "127.0.0.1:50051";
+pub const DEFAULT_LISTEN: &str = "0.0.0.0:50095";
```
kaileliu, line 10: Non-blocking follow-up: this public bind conflicts with the README guidance to keep Keystone private and use a Liaison proxy; Liaison explicitly does not proxy these account RPCs, and the Webots manifest also exposes this port for direct client access. Please align the documentation and default, and document the transport/trust boundary for session tokens if direct remote access is the intended design.

kaileliu (approved):
The identity model, persistence safeguards, and Liaison integration are coherent, and the focused CI coverage is substantial. Approving this direction; the two inline notes are non-blocking follow-ups to tighten distributed voiceprint consistency and make the network trust boundary unambiguous.

Issue #177, opened 2026-07-23
[BUG] Scene object world model retains inaccurate, stale, and off-map ghost objects


enkerewpo:
## Problem

<img width="1822" height="1100" alt="Image" src="https://github.com/user-attachments/assets/3911302d-ed5c-4885-a4b5-9f441b60e642" />

Scene's object world model is not reliable enough for object-oriented operations such as finding an object, navigating near it, or using its geometry. On the physical desktop robot we repeatedly observe:

- incorrect class labels;
- inaccurate 3D bounding boxes and point clouds;
- object centers that move between updates or appear far outside the mapped rooms;
- wall/floor detections and other ghost objects;
- objects that remain indefinitely after they are removed;
- no supported runtime way to delete, correct, or flush a bad derived object;
- invalid objects admitted during startup while the robot pose estimate or camera extrinsics are not yet trustworthy.

These failures compound: an early bad observation becomes persistent state, later detections merge into it, and operators cannot repair the result. This makes Scene unsuitable as the authoritative object state for navigation or retrieval even when its room/map functions are otherwise available.

## Current implementation findings

This is not one detector-accuracy bug. The current pipeline lacks admission, provenance, lifecycle, correction, and spatial-consistency gates.

### 1. Label and geometry quality

- The ConceptGraphs path uses a fixed default office-oriented class list. Task-relevant classes outside that list are never proposed unless the whole list is overridden through `SCENE_OPEN_VOCAB_CLASSES`.
- The RGB-only path is explicitly office-biased, limits the returned detections, and estimates distance/size rather than measuring them. It may also fall back to a zero robot pose and fixed camera assumptions. Its geometry must not be treated as navigation-grade.
- The RGB-D path back-projects depth and masks, but the current 3D box is derived from partial visible points using centroid/PCA/percentiles. Mask leakage, occlusion, and sparse depth directly perturb center and size.
- Single-frame back-projection currently skips configured DBSCAN in the main call path. Ground rejection is category/height heuristic only; there is no general wall-plane, occupancy-bound, ray-consistency, or maximum-range gate.
- Cross-class overlap/near-distance cleanup can merge detections even when their semantics disagree, allowing wall/floor leakage to contaminate a real object.

### 2. Startup pose/extrinsics pollution

[...]

PR #179, opened 2026-07-23
feat(scene): expose stable room region queries

enkerewpo:
## Summary

Add a focused `list_regions` Scene contract for room discovery and make `goal_room` fail closed when room geometry is stale. This is intentionally separate from the broader Scene object-world-model reliability work tracked in #177.

## Changes

- add `Region.msg`, `ListRegions.srv`, and `robonix/system/scene/list_regions`
- return stable `scene.room.<annotation_id>` identifiers, full polygon geometry, map ID, update time, and explicit staleness
- preserve room entries in `list_objects.v1` for backward compatibility while directing new callers to `list_regions`
- reject stale room geometry before checking live navigation geometry or returning a goal
- report an unavailable annotation store as an error instead of an ambiguous empty map
- register the new contract in native, Jetson Docker, and Jetson native Scene manifests

## Stacked review order

This PR is temporarily based on #194 (`agent/soma-footprint-scene`) so its diff contains only the room/region increment and uses the Soma-backed footprint planner. After #194 enters `dev`, retarget this PR to `dev`; do not merge this PR first.

## Validation

- real clean codegen from this branch: 42 packages, 205 messages, 111 services; all 82 generated Python modules imported successfully
- complete Scene suite: 174 passed (excluding the known macOS Bash 3 fake-Docker launcher case that runs on Linux CI)
- runtime portability audit and `git diff --check`: passed
- commit authorship: passed

The object label, point-cloud, 3D-box, association, and dropout experiments remain isolated under #177.

### DIFF
--- a/capabilities/lib/semantic_map/msg/Region.msg
+++ b/capabilities/lib/semantic_map/msg/Region.msg
@@ -1,0 +1,11 @@
+# User-authored room region anchored to the active SLAM map.
+# `id` is the stable Scene identifier accepted directly by goal_room.
+
+string id
+string kind
+string name
+float64[] points_xy
+float64 theta
+bool stale
+string stale_reason
+float64 updated_at_unix
--- a/capabilities/lib/semantic_map/srv/ListObjects.srv
+++ b/capabilities/lib/semantic_map/srv/ListObjects.srv
@@ -1,8 +1,6 @@
 # robonix/system/scene/list_objects — return every object the scene
-# registry currently believes exists. No filters, no scoping; the LLM
-# calls this every round to ground its world model and filters
-# client-side by label / distance / etc. (cheaper than baking those
-# knobs into the schema).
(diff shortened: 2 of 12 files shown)


enkerewpo (commented):
Reviewed the Region IDL/contracts, manifest declarations, MCP registration, `list_objects` v1 compatibility, stable room-ID resolution, stale-room rejection, occupancy-grid goal selection, and focused tests.

No blocking code finding from this pass. The API split is coherent: `list_objects` remains backward compatible, `list_regions` exposes full polygon/map/staleness data, and `goal_room` refuses stale geometry before navigation.

The red Rust check was not a code/test assertion: GitHub annotated that the hosted runner lost server communication while `cargo build` was in progress, with no job log produced. The failed job has been rerun. Future follow-up should add deterministic ordering and explicit malformed-polygon tests, but neither changes the current contract semantics.

PR #182, opened 2026-07-23
ci: enable required checks on dev-next

enkerewpo:
## Summary

Sync the stable CI coverage change already merged into `dev` through #181 into `dev-next`.

This makes pushes and pull requests targeting `dev-next` run:

- Rust CI and Webots smoke checks
- commit authorship checks
- docs checks

It closes the coverage gap observed after #180, where only triage jobs ran against `dev-next`.

## Validation

- all #181 checks passed, including Rust fmt/clippy/build/tests
- parsed the three workflow files as YAML
- 13 commit-authorship unit tests passed
- `git diff --check`

## Scope

CI workflow routing only; no Pilot experiment code is merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,9 @@ name: CI
 
 on:
   push:
-    branches: [main, dev, dev/**]
+    branches: [main, dev, dev-next, dev/**]
   pull_request:
-    branches: [main, dev, dev/**]
+    branches: [main, dev, dev-next, dev/**]
 
 env:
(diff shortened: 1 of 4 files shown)


enkerewpo:
Superseded by #183. The newly enabled docs check correctly exposed pre-existing CJK fixtures in `services/speech/config.spec`. Because this branch originated from `dev`, adding the existing `dev-next` file here creates divergent add/add history; #183 is based directly on current `dev-next`, carries the same CI routing, and fixes the docs violation cleanly.

PR #186, opened 2026-07-25
ci(webots): preserve spatial state across lifecycle reboot

enkerewpo:
## What changed

- preserve mapping and scene persistence across the Webots boot → shutdown → boot lifecycle pass
- reset only fixed-key agent-memory fixtures between scenario passes
- document that the second boot intentionally reuses both the running simulator and its persistent spatial state

## Why

The previous lifecycle setup kept the physical simulator at its post-scenario pose but deleted the mapping database and scene state before restarting Robonix. That mixed a moved robot with a fresh map session. In run `30148154807` attempt 2, the first 17/17 scenarios, scene-map persistence, and boot/shutdown/reboot checks passed, but the second `object_navigation` pass failed while RTAB-Map was continuously rebuilding/resizing the map and Nav2 treated the robot start as occupied.

A normal process reboot must preserve persistent spatial state. The scenario's immutable `ci_test_map` fixture is still removed explicitly before the second pass, so fixture isolation remains intact.

## Validation

- `git diff --check`
- workflow YAML parsed successfully
- lifecycle reset contract assertion confirms mapping/scene paths are preserved and agent memory is reset
- `python testing/simulate_ci.py`: 17 scenarios passed
- full Webots lifecycle and post-restart scenario suite will run in CI

Assisted-by: Codex:gpt-5.6

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1043,15 +1043,20 @@ jobs:
         # state under <manifest-dir>/rbnx-boot/ (not ~/.robonix/), that
         # Soma package children are stopped cleanly, and that the stack
         # can come back up after a clean teardown (the issue #128
-        # regression). The second boot reuses the per-run sim container;
-        # the validator does NOT start a new webots instance. CI leaves
-        # the second boot running so the next step can exercise scenarios
-        # against the restarted stack; the final cleanup step shuts it down.
+        # regression). The second boot reuses both the per-run sim container
+        # and the persistent mapping/scene state: a process restart must not
+        # silently turn into a new-map session while the physical robot stays
+        # at its post-scenario pose. CI leaves the second boot running so the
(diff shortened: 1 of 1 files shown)


enkerewpo (commented):
Review result: no blocking code findings.

The workflow change matches the intended lifecycle contract: the simulator keeps the robot at its post-scenario pose, so mapping and scene persistence must survive the Robonix process restart. Fixed-key agent-memory fixtures remain isolated, and `ci_test_map` is still removed explicitly before the second scenario pass.

Validation reviewed:

- diff and YAML structure are valid
- scenario harness simulation passes all 17 scenarios
- Rust, streaming, docs, authorship, label, and security checks are green
- the first trusted GPU dispatch did not reach checkout because an unrelated live Webots/scene session left only 1860 MiB free versus the 4096 MiB safety threshold; this is runner contention, not a code result

Full Webots execution remains required before merge.

PR #187, opened 2026-07-25
feat(navigation): add dynamic speed limit contracts

enkerewpo:
## Summary

- replace the overloaded speed RPC with three focused contracts:
  - `robonix/service/navigation/adjust_speed`
  - `robonix/service/navigation/set_speed_limit`
  - `robonix/service/navigation/get_speed_limit`
- separate relative voice-style commands from explicit policy and read-only state
- report both the effective percentage and the SI planar linear-speed ceiling in m/s
- support goal-scoped changes that restore automatically and explicit session-persistent changes

## Interface

- `adjust_speed(direction, run_id, persist)` accepts `faster`, `slower`, or `normal`
- `set_speed_limit(percentage, run_id, persist)` sets an explicit bounded percentage
- `get_speed_limit()` reports the configured maximum/default/minimum/step and current effective limit

Empty `run_id` selects the most recent active navigation run for a non-persistent mutation. `persist=true` applies the limit to the provider session instead.

The public speed ceiling is defined as planar `sqrt(vx^2 + vy^2)` in metres per second. Percentages are relative to the deployment-configured `max_linear_speed_mps`; angular constraints remain controller-owned and are not part of this API.

## Validation

- `cargo clippy --workspace --all-targets -- -D warnings`
- clean package codegen with MCP + gRPC Python import validation: 82 modules
- local docs script could not execute because macOS ships Bash 3 without `mapfile`; the Ubuntu CI docs check is authoritative

## Downstream

- provider: syswonder/service-navigation-rbnx#8
- robot configuration: syswonder/robot-agilex-ranger_mini_v3 (branch `agent/navigation-speed-config`)


enkerewpo (commented):
Review update after the linear-only correction: no blocking contract or codegen issue found. Relative commands, explicit policy, and read-only state remain distinct; the public API now reports only the effective planar linear limit in SI m/s, while angular constraints stay controller-owned. Full local fmt, clippy, workspace tests, and clean 82-module codegen validation pass.

enkerewpo:
Tested on ranger robot

PR #188, opened 2026-07-26
fix(api): allow loopback MCP host headers

Origamii520:
## Summary
- keep DNS rebinding protection enabled for loopback-bound MCP providers
- allow only the exact loopback host, its dynamic port, and the equivalent localhost spelling
- preserve wildcard/LAN behavior without adding an allowlist

This fixes local Atlas MCP requests being rejected with HTTP 421 when FastMCP receives explicit transport settings.

## Validation
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=pylib/robonix-api python3 -m pytest -q pylib/robonix-api/tests/test_provider_bind_host.py` (6 passed)
- `python3 scripts/check_commit_authorship.py --base origin/dev --head HEAD` (1 commit passed)
- `git diff --check origin/dev`

### DIFF
--- a/pylib/robonix-api/robonix_api/capability.py
+++ b/pylib/robonix-api/robonix_api/capability.py
@@ -114,6 +114,17 @@ def _resolve_provider_id(default_id: str) -> str:
     return instance_id
 
 
+def _mcp_loopback_allowed_hosts(bind_host: str) -> list[str]:
+    """Return exact/wildcard Host values accepted by a loopback MCP server."""
+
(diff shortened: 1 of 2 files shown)


Origamii520:
This loopback MCP Host compatibility fix is now ready for maintainer review. All current PR checks are green, including the human-authorship policy, Rust workspace checks, docs check, and Webots smoke test. It is a dependency of the validated Go2 local provider path and does not broaden non-loopback exposure.

enkerewpo (approved):

PR #195, opened 2026-07-27
fix(core): remove deployment-specific spatial fallbacks on dev-next

### LINKED ISSUE #193: [BUG] Core runtime contains deployment-specific robot, frame, and environment hardcodes
## Problem

Robonix core is intended to run unchanged across robots, but several generic runtime paths still contain deployment-specific assumptions about robot geometry, TF frame names, camera mounting, providers, and developer environments.

The immediate failure is Scene `goal_room`: it evaluates every robot as a TIAGo-sized disc instead of consuming the active footprint already published by Soma. This makes a quadruped report `no known free pose inside room` for valid regions. The same class of bug also exists outside Scene, so this issue tracks a repository-wide audit and guardrail rather than a Scene-only patch.

Hardcoding here means a generic core component silently invents deployment state. Constants that are part of a public protocol, explicit examples/tests/mocks, image-internal paths, and clearly named hardware adapters are not violations.

## Repository-wide audit

Audited runtime sources under `system/`, `services/`, `tools/`, `pylib/`, and public capability/contract definitions. Examples, tests, generated files, and explicit hardware adapters were inspected separately so they are not confused with portable core behavior.

### Confirmed behavior-affecting violations

| Area | Assumption | Why it is invalid |
| --- | --- | --- |
| Scene goal planning | Fixed `0.3 m` “TIAGo-sized” radius, `0.4 m` clearance, `0.5 m` object standoff, and a fixed ring search | Ignores `robonix/system/soma/footprint`; rejects valid poses for one robot and can admit unsafe poses for another |
| Scene self model | Fixed `0.6 x 0.6 x 1.5 m` robot box | Pollutes Scene relations, state, and visualization with invented embodiment geometry |
| Scene localization | Falls back to literal `map -> base_link` TF frames | Missing contracts become plausible but wrong coordinates instead of an explicit not-ready state |
| Scene metric perception | Defaults to `camera_optical_frame`, `map`, and a `1.1 m` camera height | Missing extrinsics are replaced by a guessed head-camera mount |
| Scene RGB-only perception | Can use a zero robot pose, fixed camera-axis rotation, `1.1 m` height, and emit `map` objects | Direct source of off-map, jumping, wall, and floor ghost objects tracked by #177 |
| Scene object state/persistence | `Pose3D`/`BBox3D` and restored records default to `map` | Coordinates with no proven frame are silently relabelled as map coordinates |
| Scene web UI | Builds a fixed TIAGo-like base, torso, head, and arm | Operator UI displays false robot geometry on carts, quadrupeds, and other manipulators |
| Scene -> Memory | Scene hook/watchdog publish spatial origin as literal `world` | Loses the actual `frame_id` carried by Scene objects |
| Memory | `SpatialContext` and import/build helpers default missing origin to `world` | Persists ambiguous coordinates as if their reference frame were known |
| Memory build | Searches `/home/hyl/miniconda3/envs/env_robonix/bin/uv` | A generic package build depends on one developer's home directory |
| Pilot planner prompt | Gives `tiago_camera.camera_snapshot` as a runtime provider example | Biases the model toward an implementation name that may not exist on the active robot |
| Python API guidance | Top-level example constructs a Mid-360 provider and defaults to `/scanner/cloud` | Encourages new packages to copy a device/topic assumption instead of using manifest configuration and Atlas contracts |
| Soma source tree | Uncompiled `health.rs` contains a fixed Piper body, six joints, `base_link`, and Piper fault IDs | Dead generic-looking core source is unsafe to revive or copy and has no place in the portable module tree |

The Scene perception items are also concrete root-cause inputs for #177. This issue removes invented geometry and frames; #177 will then measure and improve detection, projection, association, label quality, and lifecycle behavior separately.

### Reviewed and intentionally allowed

- `examples/**`, `testing/**`, workflow fixtures, and explicit Webots/TIAGo demonstrations may name their test robot.
- Explicit hardware adapters such as `vitals/scripts/piper_bridge.py`, Koch/Piper configuration, and vendor-specific package manifests may contain hardware-specific joints, ports, or models.
- Standard local control-plane defaults such as Atlas loopback addresses are protocol defaults, not robot embodiment assumptions.
- Container image paths such as `/opt/models` and `/opt/ros` are owned by the image definition and configurable at the package boundary.
- Stable contract IDs, lifecycle values, protobuf fields, and test fixture provider names are not deployment leakage.
- The default open-vocabulary class profile is environment-specific, but changing model vocabulary belongs to the measured perception work in #177 rather than this plumbing patch.

### Geometry and planning

- Resolve `robonix/system/soma/footprint` through Atlas and cache the validated footprint polygon, base frame, inscribed radius, and circumscribed radius.
- Use the polygon for room-boundary and occupancy tests. Derive object standoff from target geometry plus the current footprint, rather than a robot-name default.
- Fail closed with an explicit `Soma footprint unavailable` result while geometry is unknown; never substitute simulator geometry.
- Represent the Scene self object and web view from the active footprint. Do not invent unknown Z dimensions.

### Coordinate frames and readiness

- Consume standard map pose/odometry and camera extrinsics contracts, preserving the frame IDs supplied by those providers.
- Do not transform or admit spatial objects until pose and extrinsics are available and temporally usable.
- Missing frames remain unknown and produce an explicit not-ready/error state; they must never become `map`, `world`, `base_link`, a zero pose, or a guessed camera mount.
- Carry the actual common frame into Memory. Reject mixed-frame snapshots unless they have been transformed deliberately.

### Environment and model independence

- Discover build tools from the package venv, `PATH`, or documented configuration; remove developer-home absolute paths.
- Keep robot/provider names out of generic LLM prompts and generic API examples. Require exact names discovered through Atlas/configuration.
- Remove dead embodiment-specific core sources or move behavior into an explicitly named adapter package if it is still required.

## Prevention

Add a focused source audit that runs in CI over generic runtime directories. It should reject:

- known robot/provider names in generic runtime prompts and behavior;
- user-home absolute paths;
- known fixed geometry fallbacks;
- silent spatial defaults such as guessed robot/camera frames or zero poses.

The allowlist must be narrow and path-based for explicit examples, tests, mocks, generated output, protocol defaults, and hardware adapters. The report should identify the file, line, matched rule, and remediation; it must not turn legitimate adapters into false failures.

## Acceptance criteria

- The audit has a reviewed finding for every match in `system/`, `services/`, `tools/`, `pylib/`, and public contracts, with allowed matches documented above.
- A Webots TIAGo fixture and a non-TIAGo Soma fixture produce different planning geometry from the same Scene code.
- `goal_room` and `goal_near` use the footprint returned by Soma and contain no robot-specific geometry constants.
- When footprint, pose, or camera extrinsics are unavailable, Scene reports not-ready and admits no guessed map-frame objects; it recovers when the standard contracts become available.
- Scene and Memory preserve the real coordinate frame end to end and reject ambiguous/mixed-frame spatial data.
- No generic build path references a developer home directory.
- Generic runtime prompts and public API guidance contain no robot/provider-specific default.
- The source audit, focused Python tests, Rust fmt/clippy/build/tests, codegen/import checks, and Webots integration pass on both `dev` and `dev-next`.

## Delivery

Deliver equivalent fixes to `dev` and `dev-next` in separate PRs. Keep the subsequent metric/label/3D-box/association/dropout experiments for #177 in a separate branch and PR so their Webots A/B results remain reviewable.

enkerewpo:
## What changed

- resolve the robot footprint and base frame from the standard Soma footprint capability instead of Webots- or robot-specific constants
- make `goal_room` validate the complete transformed footprint against room polygons and occupancy cells; make `goal_near` derive clearance from live robot geometry
- keep the dev-next TF-first Scene experiment, but require explicit/header-derived world, body, and camera frame endpoints and a fresh pose before publishing spatial detections
- remove camera-height, zero-pose, hard-coded frame, developer-path, and provider-name fallbacks from generic runtime paths
- add a portability audit to CI and document the standard Soma/Scene contract boundary

## Root cause

Generic runtime code had accumulated deployment assumptions such as Webots footprint dimensions, conventional ROS frame names, and guessed camera geometry. Those guesses could produce apparently valid goals or object coordinates for an incompatible robot. dev-next also had a TF-first experiment whose compatibility fallback could silently compose an incomplete frame chain.

## dev-next adaptation

This is not a wholesale merge from `dev`. It preserves dev-next lifecycle and TF-first perception behavior, and adapts only the portability/geometry fix to that architecture. Modules removed on dev-next (including Memory integration) remain removed.

## Validation

- real `rbnx codegen --mcp --ros2 --clean`: 43 packages, 207 messages, 106 services
- imported all 127 generated protobuf/MCP Python modules
- Scene tests: 203 passed
- `cargo clippy --workspace --tests -- -D warnings`
- `cargo build --workspace`
- Webots streaming smoke and LaserScan normalization tests
- IDL comment, Python API discovery, commit authorship, diff, compile, and runtime portability checks

Fixes #193


system/scene/scene_service/mcp_tools.py
```diff
@@ -527,7 +484,19 @@ async def goal_room(req: GoalRoom_Request) -> GoalRoom_Response:
...
+    if not grid_frame:
+        return GoalRoom_Response(
+            reachable=False,
+            x=0.0,
+            y=0.0,
+            yaw=0.0,
+            reason="occupancy_grid frame is unknown",
+        )
+    room_yaw = float(room.theta or 0.0)
+    found = room_goal(msg, room.points, footprint, yaw=room_yaw)
```
enkerewpo: [P1] Search a valid robot heading instead of forcing room yaw to zero. Room annotations cannot carry theta, so this always selects 0 rad. An anisotropic Go2 footprint (0.8 x 0.2 m) in a 0.36 x 1.0 m room returns unreachable at yaw=0 even though yaw=pi/2 yields a valid footprint-safe pose; this recreates a false-unreachable failure for non-circular robots.

system/scene/scene_service/robot_geometry.py
```diff
@@ -0,0 +1,186 @@
...
+    *,
+    fetcher: Callable[[], Optional[RobotFootprint]] = fetch_soma_footprint,
+) -> None:
+    """Continuously reconcile Soma geometry without blocking Scene startup."""
+    retry_s = 1.0
+    refresh_s = 30.0
+    last_error = ""
+    while True:
+        try:
+            footprint = await asyncio.to_thread(fetcher)
```
enkerewpo, line 159: [P1] Clear or expire the cached footprint when refresh fails. After one successful fetch, None/exception paths leave RobotGeometryState.current() populated forever, so goal_room/goal_near continue planning with stale geometry after Soma disappears or the deployed robot changes, despite the log claiming fail-closed behavior.

system/scene/scene_service/service.py
```diff
@@ -718,7 +784,18 @@ def _rgb_jpeg() -> Optional[bytes]:
         if hub.has("intrinsics"):
             msg, stamp, _ = hub.latest("intrinsics")
             if msg is not None and stamp > 0.0:
-                vlm_intrinsics = _cam_info_to_intrinsics(msg)
+                info_frame = str(
+                    getattr(getattr(msg, "header", None), "frame_id", "") or ""
+                ).strip()
+                if info_frame == _active_camera_frame():
```
enkerewpo: [P1] Resolve VLM intrinsics dynamically rather than once at startup. The ROS subscription may exist before its first CameraInfo sample, leaving vlm_intrinsics=None permanently; VLMObjectDetector then rejects every projection even after the contract starts publishing, so the documented recovery behavior never occurs.

PR #196, opened 2026-07-27
fix(scene): make object lifecycle observation-aware

enkerewpo:
## Summary

This is the first incremental fix for #177. It separates current-frame observations from the persistent ConceptGraphs map so historical objects can no longer refresh themselves forever.

- refreshes `last_seen` and `observation_count` only for UUIDs carrying current-frame evidence;
- treats absence as negative evidence only on a healthy RGB-D frame when the old location is in view and depth shows clear space behind it;
- marks objects missing after a configurable number of visible misses, then prunes them after a configurable TTL;
- preserves stable IDs when objects reappear before the TTL;
- clears the ConceptGraphs map, UUID bindings, registry objects, and Scene Graph caches together on map epoch changes and map loads;
- moves lifecycle thresholds to `system.scene.config.perception` with SI units and documents their semantics;
- adds a live Webots Supervisor fixture that spawns, removes, respawns, and finally expires a real chair through the complete Scene pipeline.

This PR is intentionally stacked on #195 (`agent/dev-next-portability`) because #177 work targets the experimental `dev-next` Scene pipeline. After #195 is reviewed/merged, this PR should be retargeted to `dev-next`.

## Why

The previous projection path serialized the entire historical `MapObjectList` after every tick and treated every retained entry as a fresh positive observation. Consequently `last_seen` advanced forever, `missing` was cleared forever, and dropout/TTL could never remove a moved object.

The new lifecycle distinguishes:

1. a positive current-frame match;
2. a visibility-verified miss;
3. unknown state caused by occlusion, out-of-FOV, sensor/model failure, or unavailable spatial transform.

Only (1) refreshes an object, and only repeated (2) can mark it missing.

## Validation

- `211 passed` across `system/scene/tests` using clean generated Scene protobuf/MCP modules from the #195 base;
- focused lifecycle/map/identity/persistence suite: `45 passed, 23 skipped` (Milvus-Lite skips on macOS are expected);
- all 15 integration scenarios compile and pass the offline harness simulation;
- Python compile check passed for every modified Scene module and the live verifier;
- `git diff --check` passed.

The latest-head Webots integration run exercises the new `scene_object_dropout` scenario through the trusted Robonix CI path. It must report fixture appearance, visible-miss transition, same-ID reappearance, and TTL hard deletion before this PR is considered verified.

## Scope left for follow-up PRs

This PR repairs readiness/epoch reset and dropout state semantics. It does not claim to finish #177. Spatial admission/plane filtering, multi-view geometry and bbox uncertainty, label posterior/association, and supported CRUD/flush contracts remain separate, measured follow-ups.

Refs #177

### DIFF
--- a/examples/webots/robonix_manifest.yaml
+++ b/examples/webots/robonix_manifest.yaml
@@ -29,7 +29,16 @@ system:
   vitals:
     listen: 0.0.0.0:50093
   # soma+scene layer
-  scene: {}
+  scene:
+    config:
+      perception:
+        # Metric SI units. A removed object must be absent in three healthy,
+        # depth-verified frames before it becomes missing. Sensor/model
+        # outages and occlusion do not count as negative observations.
+        period_s: 0.6
+        visible_miss_frames: 3
+        visibility_depth_margin_m: 0.20
+        object_ttl_s: 30.0
   # execution layer
   executor:
     # robonix-client polls Executor for authoritative active RTDL plans.
--- a/system/scene/README.md
+++ b/system/scene/README.md
@@ -48,7 +48,7 @@ The metric pipeline is ConceptGraphs-style per-frame perception with 4 stages:
    * **Distance gate** — centroid > 1.5 m apart → never merge (kills "9 × 5 m bbox spanning the room" failure).
    * **Same-class gate** — different YOLO class names → never merge (kills "potted_plant on cabinet collapses to one record").
    * **Threshold** — agg_sim < 0.55 → spawn new object instead.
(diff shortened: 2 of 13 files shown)


enkerewpo:
@robonix-ci test

enkerewpo:
@robonix-ci test

enkerewpo:
Superseded by the consolidated Scene reliability PR #199. All commits and changes from this PR are ancestors of #199's current head; this closes only the redundant review wrapper, and no code is being dropped or merged.

PR #197, opened 2026-07-27
fix(scene): improve RGB-D geometry and spatial admission

enkerewpo:
## Summary

This is the second incremental #177 change, stacked on #196. It improves metric RGB-D point clouds and 3D boxes without mixing label/association changes into the same review.

- erodes one configurable SAM boundary pixel, while automatically retaining the original mask when erosion would destroy a small valid object;
- applies sensor-range and robust median/MAD depth filtering before backprojection;
- enables the already-configured per-frame ConceptGraphs DBSCAN path;
- withholds authoritative objects until a same-frame occupancy grid exists;
- rejects point clouds outside the current (possibly rotated) occupancy-grid bounds;
- rejects per-frame hypotheses with an invalid or deployment-oversized robust bbox;
- derives object center, yaw, and extents from trimmed PCA plus configurable quantiles instead of a mean shifted by outliers;
- publishes cumulative admission counters through `/api/state.perception_quality`;
- documents every parameter under `system.scene.config.perception.geometry`, with SI units and semantics.

## Configuration contract

All metric distances are metres. `bbox_*_percentile` uses 0–100 percent; `map_max_outside_fraction` uses 0–1. The Webots deployment opts into the documented defaults explicitly so CI exercises the public Driver contract rather than private environment variables.

## Validation

- Scene test suite: `216 passed`;
- new pure regressions cover mask-edge leakage, small-object erosion fallback, rotated occupancy grids, far ghost rejection, frame mismatch, and bbox resistance to a large depth spike;
- all 16 integration scenarios compile and pass offline harness simulation;
- runtime portability audit passed;
- Python compile and `git diff --check` passed.

The new Webots `scene_object_quality` scenario checks every live non-robot object against the real occupancy map, requires finite positive bboxes below the configured maximum, rejects background labels, and records the admission counters. A trusted real Webots run will be attached before review completion.

## Stack

- base: #196 (`agent/scene-object-lifecycle-177`)
- this PR: spatial/point-cloud/bbox quality only
- follow-up: label evidence and association
- follow-up: epoch-checked object CRUD/flush

After #196 is merged/retargeted, this PR should be retargeted accordingly. Do not merge the stack out of order.

Refs #177

### DIFF
--- a/examples/webots/robonix_manifest.yaml
+++ b/examples/webots/robonix_manifest.yaml
@@ -39,6 +39,22 @@ system:
         visible_miss_frames: 3
         visibility_depth_margin_m: 0.20
         object_ttl_s: 30.0
+        geometry:
+          # Reject RGB-D points outside the physical sensor range, remove
+          # one-pixel SAM boundary leakage, denoise each frame, and admit only
+          # point clouds that fit the current map grid.
+          mask_erosion_px: 1
+          min_depth_m: 0.15
+          max_depth_m: 6.0
+          depth_mad_scale: 3.5
+          depth_min_band_m: 0.12
+          frame_dbscan: true
+          require_occupancy_bounds: true
+          map_bounds_margin_m: 0.25
+          map_max_outside_fraction: 0.20
+          bbox_low_percentile: 5.0
+          bbox_high_percentile: 95.0
+          max_bbox_extent_m: 3.0
   # execution layer
   executor:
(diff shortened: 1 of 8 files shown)


enkerewpo:
@robonix-ci test

enkerewpo:
Superseded by the consolidated Scene reliability PR #199. All commits and changes from this PR are ancestors of #199's current head; this closes only the redundant review wrapper, and no code is being dropped or merged.

PR #199, opened 2026-07-27
feat(scene): make derived object state correctable and recoverable

enkerewpo:
## Summary

Scene's derived object world state becomes correctable and recoverable, with ConceptGraphs still the perception backbone. Full and lite deployment profiles become reproducible for x86 simulation and Jetson targets.

Rebased onto `dev@ab765d12`.

## Product changes

- add public list, relabel, geometry-update, delete, and flush contracts for derived Scene objects;
- keep operator edits separate from model observations, and revalidate background cleanup before applying mutations;
- gate RGB-D ingestion on usable transform/map evidence, and expose quality/lifecycle diagnostics;
- preserve stable IDs, observation history, bounded association, duplicate cleanup, persistence, and registry projection;
- provide reproducible full/lite x86 images plus Jetson Docker/native full/lite manifests;
- preserve current `dev` behaviour such as `list_regions`, Soma-provided robot footprint, and the shared lifecycle Driver configuration channel.

### Objects that are removed now disappear

On `dev` the ConceptGraphs `MapObjectList` only ever grew. The registry's eviction rule is "the source uuid is gone from the map", and the uuid never went away — so an object carried out of the room stayed in the map forever. The fix is a closed loop:

| Step | Where | What it does |
|---|---|---|
| 1 | `_visible_missing_uuids` | projects the stored cloud into the current depth image and samples it; a miss counts only where the measured surface is materially **behind** the stored object. Occlusion, similar depth, and invalid depth are all "unknown" and never delete |
| 2 | `_project_to_registry` | `visible_miss_threshold` (3) consecutive misses → `missing = True`, `missing_reason = "visible_absence"` |
| 3 | `prune_expired` | after `object_ttl_s` (30 s) the record is deleted and its uuid recorded as expired |
| 4 | `_purge_expired_map_objects_locked` | drops those uuids from the ConceptGraphs map itself — the step `dev` was missing |

The design is deliberately asymmetric: positive evidence is required to delete, so the failure mode is a stale object rather than a deleted real one. `tests/test_visible_absence.py` locks that rule down — occlusion, a similar depth reading, invalid depth, and out-of-frame must all decline to delete — and is mutation-checked: counting occlusion as clear fails three of its tests, and disabling the step-4 purge (i.e. restoring `dev` behaviour) fails another.

### What actually improves accuracy

Active by default: confirmation gating (an object needs ≥2 independent frames before it is reported, which removes single-frame ghosts), confidence-weighted label voting with switch hysteresis, globally one-to-one association, an identity gate requiring independent spatial, visual, and extent evidence, floor-noise rejection, occupancy-bounds rejection, same-frame duplicate collapse, and periodic cleanup that revalidates against live state before applying.

**Three mechanisms ship inert and are excluded from the numbers below**: CLIP label rerank, planar surface snapping, and stationary tile refinement. All three are scoped only from Driver config, and nothing in the repository scopes them, so they never ran — including during measurement. `perception_tuning.py` now carries validated candidate scopes (`CANDIDATE_CLIP_RERANK_GROUPS` and friends) and the README documents how to adopt one, but the defaults stay empty until they are measured. Turning any of them on invalidates the table below and needs a re-run.

## Slimming pass

The first three commits were reduced by a further **1186 lines** net (+1635 / −2821 over 15 files), with no behaviour change. In production perception code the reduction is larger, because part of the +1635 is new tests and docs:

| File | Before | After |
|---|---:|---:|
| `ingest/perception_concept_graphs.py` | 9947 | **8512** |
| `service.py` | 2923 | **1935** |
| `ingest/perception_tuning.py` (new) | — | 965 |
| **net production code** | | **−1458** |

What went, inside that net figure — these overlap in one commit, so they do not sum:

- perception config plumbing, collapsed into `PerceptionTuning`;
- dead code, ~141 lines: `_merge_overlap_no_p3d` (self-labelled UNUSED), `_maybe_periodic_cleanup` (documented as a hook "retained for focused unit tests" that no test calls), and two orphaned shims;
- the co-observed merge diagnostics, ~500 lines across three methods — reporting only, never a merge decision, but an O(N²) geometry pass plus two 256-frame history scans per pair on every periodic cleanup. The cheap counters from the merge's own decision path stay;
- `quality_metrics`, 380 → 111 lines, rebuilt on the tuning record instead of hand-written `getattr(self, "_x", default)` per knob.

Every perception knob used to be spelled three times — extracted and range-checked in `service.py`, forwarded as one of 102 keyword arguments, then re-clamped in `ConceptGraphsDetector.__init__` — so the three copies were free to disagree, and `service.py` rejected values the detector would silently clamp. Each knob is now declared once, validated once, and read once. All 87 defaults were diffed against the previous keyword-argument defaults and are identical.

Two intentional behaviour changes come with that: an out-of-range config value now **fails the boot** naming its `perception.…` path instead of being clamped into range, and a self-referential label alias is rejected instead of silently dropped.

## Fixes for review comments on this PR

- **`grpcio==1.80.0` made the uv workspace unsatisfiable** against `services/memory` (`>=1.81.1`). Root cause was wider than the pin: this branch changed scene's dependencies and never re-locked, so `uv.lock` disagreed with the manifest. Both are now floors — generated `_pb2_grpc.py`/`_pb2.py` enforce a *minimum* runtime, and reproducibility belongs to the generator, which stays pinned exactly in its isolated codegen venv. `verify_python_codegen.py` gained `>=` support for the same reason: after the floor, a host carrying grpcio 1.81.1 is correct, and asserting equality there would have failed the build it was meant to protect.
- **PEP 668 blocked the jetson-native lite build.** Lite needs no JetPack torch, so it now installs into its own `rbnx-build/native-venv` (`--system-site-packages`, keeping host rclpy visible); `start_native.sh` picks it up automatically. The metric profiles keep the venv-free `pip --user` path, because a venv cannot see a JetPack torch installed in `~/.local`.

## Other defects found and fixed

- The README's env table and troubleshooting entries still prescribed the class-gated identity knobs (`SCENE_CG_CROSS_CLASS_*`, `SCENE_CG_SAME_CLASS_*`, `SCENE_CG_MERGE_CLASS_GROUPS`) as the cure for duplicate records and label flicker. This branch turned them into ignored no-ops, so an operator following the README would change nothing and conclude the symptom was unfixable.
- `tests/test_cg_kernels.py` imported `torch` at module scope, so collection of the **whole** suite failed on any host without it — including the lite profile this branch adds.
- `tests/test_identity.py`'s `__main__` block called a test the branch had removed; running the file directly raised `NameError`.
- `test_periodic_cleanup_background` and `test_persistence` built detectors via `__new__` plus ~40 hand-set attributes, and had already drifted: they assigned three attributes the detector no longer reads. They now construct through `__init__`, so a rename fails loudly.
- `test_build_offline_contract.py` was deleted rather than updated when the image moved to baked safetensors. Its subject — no registry-backed BuildKit frontend, no network weight fetch inside the build, resumable host-side downloads — is still a contract this branch claims to uphold. Restored, extended to `Dockerfile.lite`, and reduced from 193 lines to static assertions.
- Three visibility knobs (`visibility_upper_sample_fraction`, `visibility_min_depth_margin_m`, `visibility_extent_margin_scale`) had neither a config path nor a range check, despite governing how much evidence is required before Scene deletes an object.
- `verify_python_codegen.py`'s success path raised `NameError` — found by running the codegen on a real Linux host, which is exactly the run that was previously only claimed.

## Validation

Validated on exact head `ec9d42f5` (Webots run was on `daf5b725`, which differs only by the added `test_visible_absence.py`), based on `dev@ab765d12873c551ba60708e08307237baa911129`.

Run on x86_64 Linux (Debian 13, Python 3.11 via uv), from a clean clone of this branch, with `rbnx`/`robonix-codegen` built from this branch's own sources:

| Check | Result |
|---|---|
| `scripts/run_python_codegen.sh --mcp --ros2 --clean` | **82 generated Python modules validated**; ROS 2 overlay 42 packages / 205 msg / 118 srv; verifier reports `protobuf=6.33.6 grpcio-tools=1.76.0 grpcio=1.80.0` |
| Scene test suite | **237 passed, 1 skipped** (the skip is `test_cg_kernels` on a torch-less venv — the new guard working) |
| `uv lock --check` | clean (was stale on the previous head) |
| `uv sync --package robonix-memgraph` | **exit 0** — the exact repro that failed before |
| `uv sync --package robonix-system-memory` | exit 0 |
| `uv sync --package robonix-system-scene` | exit 0, resolving **grpcio 1.81.1** — the version the old exact pin made unreachable, and scene imports fine on it |
| `scripts/check_commit_authorship.py` | 11 commits passed |
| `cargo build --release -p robonix-cli -p robonix-codegen` | clean; the whole run used tooling built from this branch |

### Webots end-to-end (office.wbt, Tiago, RTX 2060)

Full `rbnx boot` of `examples/webots` — sim, chassis, camera, lidar, mapping, nav2, explore, scene — with the scene container running **this branch's source** (`start.sh` bind-mounts the package), then a ~130 s in-place panorama and a settle window.

| | |
|---|---|
| Derived objects built | **15** — cabinet ×3, television ×3, lamp ×2, chair ×2, potted_plant, sink, couch, table, keyboard |
| Detection funnel | 77 → **178 accepted** frame detections; 19 depth-rejected masks |
| Association | 546 candidate pairs → 110 kept; 12 rejected by the one-to-one gate; 14 same-frame duplicate merges |
| Confirmation gating | 14 confirmed vs 6 unconfirmed candidates |
| Background periodic cleanup | 8 scheduled / 7 completed / 5 applied, **0 failures, 0 stale-snapshot discards** |
| Registry projection | **0 failures** over 14180 healthy frames |
| Transform chain | 14087 observation-transform evidence records; RGB↔depth skew 0.02 s |
| `perception_quality.tuning` | all **88 knobs** reported live, confirming the three correction mechanisms are genuinely inert in a stock deployment (`clip_rerank_groups=[]`, `surface_snap_labels=[]`, `stationary_refinement=false`) |

This exercises the refactor across process boundaries — Driver config → `PerceptionTuning` → detector → registry → web API — which unit tests cannot. The background cleanup transaction, the part most heavily restructured, ran seven times without a failure or a stale-generation discard.

No object was spuriously marked absent over 14180 frames in a static scene, which is the safety half of the deletion path. The other half — an object actually removed from the world disappearing — is covered by `tests/test_visible_absence.py` rather than staged in the simulator; see below.

### Not validated by this PR

- **No fresh Frozen Office measurement.** The accuracy table below was measured on the pre-slim tree. The Webots run above proves the pipeline works end to end after slimming, but it is not a re-measurement of F1 or label accuracy. The slimming's neutrality rests on the per-knob default diff plus the unit suite.
- **No live object-removal scenario.** Removing a Webots node mid-run was not staged; the deletion chain is covered by unit tests instead, mutation-checked against reintroducing the `dev` behaviour.
- **No Jetson hardware.** The native lite venv change is reasoned and syntax-checked, not run on a board.

### Frozen Office protocol (measured on the pre-slim tree)

All rows use the same 120 s panorama path, 20 s settle window, FP32 full profile, TensorRT disabled, and the same frozen 23-object visible truth set. Values are three-run medians.

| System | Detection F1 | Label accuracy | Joint semantic F1 |
|---|---:|---:|---:|
| current stable-`dev` control | 0.5405 | — | — |
| pre-slim candidate | 0.8500 | 0.8235 | 0.6829 |
| exact slim head | 0.7805 | 0.9375 | 0.7317 |

`Joint semantic F1` requires a prediction to be both spatially matched and label-correct. It is included because the detection-only matcher can count semantically wrong assignments as true positives: observed examples include a `cabinet` matched to `window(B)6` at about 1.40 m centre error, and a `couch` matched to `Cabinet#1`. The slim head therefore has lower detection-only F1 than the pre-slim samples but higher label accuracy and higher joint semantic F1.

Note that the `cabinet`/`window` case is an **association** failure, not a label failure — the kind of thing `perception.association.*` governs, not the CLIP rerank.

These are preliminary engineering results for a system still under optimization; they are not final product claims.

## Deliberately excluded

- Webots/Nav/SLAM/Mapping behaviour changes;
- CI, `rbnx`, or robot-deployment changes;
- benchmark harnesses, result artifacts, report sources, and exploratory sweep code.

## Review guide

Eleven commits, grouped by review surface:

1. `feat(scene)` — object-state contracts, lifecycle, persistence, association, runtime behaviour
2. `build(scene)` — full/lite and x86/Jetson deployment profiles
3. `test(scene)` — focused regression coverage
4. `refactor(scene)` — the tuning record, dead code, and diagnostics removal (the bulk of the slimming; no behaviour change)
5. `fix(scene)` — grpcio/protobuf floors and the lockfile
6. `fix(scene)` — jetson-native lite venv
7. `docs(scene)` — README corrections
8. `test(scene)` — fixtures through the real initialiser
9. `fix(scene)` — visibility knob validation
10. `fix(scene)` — codegen verifier success path
11. `test(scene)` — visible-absence deletion coverage


enkerewpo:
CI clarification: run [30291204773](https://github.com/syswonder/robonix/actions/runs/30291204773) did **not** execute Webots or any scenario. It failed at the pre-checkout GPU-capacity gate (762 MiB free; 4096 MiB required). The bot's 17/17 table came from stale self-hosted-runner artifacts and must not be treated as validation evidence; this branch currently defines 18 scenarios. The reporting/isolation defect is tracked in #200 with draft fix #201. Real Webots validation for this Scene stack remains pending.

enkerewpo:
### Workspace dependency conflict: `grpcio==1.80.0` in scene vs `grpcio>=1.81.1` in services/memory

Found while building this branch (`9b51d0e`) on a Jetson AGX Thor (aarch64, Python 3.12, ROS 2 jazzy) to use the new `lite` perception profile for EmbodiedBench benchmarking.

`rbnx build -p services/memsearch` and `-p services/memory` both fail during `uv` resolution:

```
× No solution found when resolving dependencies for split (markers:
│ python_full_version >= '3.14' and platform_machine == 'aarch64' and
│ sys_platform == 'win32'):
╰─▶ Because robonix-memgraph depends on grpcio>=1.81.1 and
    robonix-system-scene depends on grpcio==1.80.0, we can conclude that
    robonix-memgraph and robonix-system-scene are incompatible.
    And because your workspace requires robonix-memgraph and ...
```

**Root cause**

- `system/scene/pyproject.toml:39` → `"grpcio==1.80.0"` (exact pin, introduced by this PR)
- `services/memory/pyproject.toml:16` → `"grpcio>=1.81.1"`

Both are `uv` workspace members, so the exact pin makes the workspace unsatisfiable. `uv` resolves every marker split, so this is not specific to the odd split quoted above (that combination is itself unsatisfiable) — it fails universally, not just on aarch64.

**Impact**

Any `rbnx build` of a workspace member that pulls the workspace resolution fails, so `services/memsearch` and `services/memory` cannot be built on this branch. On `dev` (`35e6bff`) both build fine.

**Reproduce**

```bash
git checkout 9b51d0ea04b8f3762e8f67a74e73c9d862a280c2
rbnx build -p services/memory
```

**Suggested fix**

Relax the scene pin to a floor unless 1.80.0 is required for a specific reason:

```toml
-    "grpcio==1.80.0",
+    "grpcio>=1.80.0",
```

Locally that single change makes both `services/memsearch` and `services/memory` build cleanly (`✓ Package ... build finished`), and scene still builds and imports fine. If the exact pin is deliberate, the alternative is to align `services/memory` down to the same version, but a floor seems preferable since nothing in scene appears to depend on 1.80-specific behaviour.

---

Separately, and not a bug — just confirming the `lite` profile works as documented: `package_manifest.jetson-native-lite.yaml` is exactly what was needed for a Jetson host with no JetPack torch installed, and `resolve_perception_profile` returning `enabled: False` for `lite` is a much cleaner way to run Scene without object perception than disabling a detector after the fact. `tests/test_perception_profiles.py` passes (4 passed).

One note on the native lite path: `RBNX_BUILD_TARGET=jetson-native ROBONIX_SCENE_BUILD_PROFILE=lite` still runs `pip install --user`, which is blocked by PEP 668 on a Debian/Ubuntu system Python:

```
error: externally-managed-environment
```

so on such hosts the native lite target can't complete without `--break-system-packages` (which I'd rather not do on a shared dev board). Using a venv for the lite deps, or documenting `jetson-docker-lite` as the option for PEP 668 hosts, would help.

HeartLinked (changes requested):
Reviewed the full diff against dev@ab765d12. Impressive scope and the deletion path's fail-safe design is right. Four things I'd like addressed before merge:

1. Vocabulary config silently disables all perception. PerceptionTuning accepts and validates perception.vocabulary.classes / additional_classes, and SCENE_OPEN_VOCAB_CLASSES is still read — but any value differing from the baked checkpoint makes _try_load_models raise, which is caught into a log.warning and returns (None,)*8. Scene still boots, reports ACTIVE, and serves an empty object list forever. Please validate the resolved vocabulary against baked_yolo_metadata() at boot and fail loudly (consistent with this PR's own "reject, don't clamp" policy). README:50 and README:301 still document the env var as a runtime override.

2. Operator corrections aren't protected from the visible-absence → TTL delete path. The periodic-cleanup paths all guard operator-owned objects; the visible-miss loop (_apply_snapshot) and prune_expired only guard is_robot. Geometry correction is the worst case: the override leaves the CG cloud at the old location, and that's exactly what _visible_missing_uuids samples — so relocating a box makes its old footprint read as clear absence and the correction is hard-deleted after object_ttl_s.

3. A single label edit wipes the whole map's persisted caption/relation cache. ObjectMutationCoordinator calls scene_graph_store.clear_derived_state() on every mutation, which clears both caches and flushes to disk. The caches are per-object/per-pair keyed and prune_relations already exists — please invalidate narrowly and reserve the full clear for flush.

4. Please disclose the ConceptGraphs replacement. The summary says ConceptGraphs is still the backbone, but the branch drops the upstream install (plus Open3D/faiss/supervision/hydra) and replaces the SLAM kernels with 760 lines of new in-repo NumPy (cg_kernels.py). That's the largest correctness risk here and it isn't in the review guide. It also means the dev 0.5405 → 0.85 comparison mixes two variables. Upstream is MIT — worth confirming attribution on the new file.

Smaller items: periodic_cleanup_discarded_stale only counts the _stop branch, so "0 stale-snapshot discards" doesn't mean what the description implies; /api/state now inlines quality_metrics() and iterates _map_objects without the inference lock; operator labels are silently rewritten by _canon_class (desk → table) with no mention in the .srv docs; pyproject.toml still declares open3d; ruff --select F reports 17 dead imports (CI only gates Rust).

enkerewpo:
Thanks for the thorough pass — all four findings verified against the branch, and they changed the plan for this PR. Point-by-point:

1. **Vocabulary silently disabling perception — confirmed.** The mismatch raises inside `_try_load_models`'s try block (`perception_concept_graphs.py:2171`), gets swallowed into `(None,)*8`, and Scene stays ACTIVE serving an empty list with only a warning.
2. **Operator corrections unprotected on the visible-miss → TTL path — confirmed.** `prune_expired` guards only `is_robot`/`restored`; a geometry correction leaves the CG cloud at the old footprint, which is exactly what `_visible_missing_uuids` samples.
3. **Whole-map cache wipe per edit — confirmed.** All three mutation paths call `clear_derived_state()`, which clears both caches and flushes.
4. **The ConceptGraphs replacement — confirmed, and it decided the outcome.** `cg_kernels.py` (760 lines) plus the dropped upstream install landed inside a 13.5k-line commit with no mention in the message. We also fact-checked the motive: `open3d>=0.17,<0.19` ships manylinux aarch64 wheels for exactly the pinned range — dev's `Dockerfile.jetson` installs upstream concept-graphs as-is, so the aarch64 justification doesn't hold. ConceptGraphs is this project's perception core; replacing it was never on the table.

**Resolution: this PR is superseded by #234**, a rebuild on `dev` that keeps upstream ConceptGraphs untouched and ports what this branch got right — the correction contracts verbatim, and the visible-absence decision function unchanged (your read on its fail-safe design matched ours; it was the part worth keeping). #234 also fixes 2 and 3 structurally (eviction immunity at the registry level; per-object cache invalidation with the full clear reserved for flush), stores operator labels verbatim (the `desk` → `table` rewrite you flagged in the smaller items), and makes 1 moot by not porting the baked-checkpoint machinery. It is +1,843/−7 over 20 files against this branch's +15,792/−2,089 over 69.

This PR stays draft until #234 lands, then closes. The 0.5405 → 0.85 comparison will be re-measured on #234's tree with the same benchmark harness (now version-controlled on `scene-benchmark-harness`) so the number no longer mixes two variables.

enkerewpo:
Superseded by #234 (rebuilt on dev with upstream ConceptGraphs intact). The branch stays available for reference; the review findings and their resolutions are recorded above.

PR #198, opened 2026-07-27
fix(scene): stabilize object labels and association

enkerewpo:
## Summary

- make the YOLO-World vocabulary deployment-configurable through `system.scene.config.perception.vocabulary`
- accumulate confidence-weighted label evidence per physical-object track and expose provisional/confidence/candidate diagnostics
- require repeated support, winner share, and switch margin before changing a stable label
- keep provisional labels out of `navigation_grade` object use
- disable unsafe cross-class proximity and geometric merging by default
- allow only deployment-reviewed `confusable_class_groups`, with the legacy unrestricted mode behind an explicit compatibility switch
- add a live Webots label-stability gate and document every parameter and unit

## Root cause

Scene previously treated one frame's YOLO-World class as the object label. It also bypassed the class gate for nearby centroids and periodically force-merged overlapping geometry regardless of class. That made transient classification errors persistent and could merge distinct co-located objects such as an item on furniture. There was no track-level evidence, convergence state, or deployment vocabulary contract.

## Validation

- `222 passed` — complete `system/scene/tests`
- `17 passed` — Webots manifest configuration tests
- `17 scenario(s)` — offline scenario simulation
- runtime portability audit passed
- Python compile and `git diff --check` passed
- live Webots ChatOps requested after this draft is opened

This is phase 3 of #177 and is intentionally stacked on #197 (`agent/scene-spatial-quality-177`). It should be retargeted to `dev-next` after the preceding PRs land. Do not merge before the live Webots label-stability result is reviewed.

PR #202, opened 2026-07-27
test(scene): measure Webots object ground truth

enkerewpo:
## Summary

This adds a fifth, validation-only layer on top of the #196–#199 Scene stack. The previous `scene_object_quality` scenario checked map bounds, finite box dimensions, and background labels, but it did not prove that a detected object had the correct class, center, 3D box, or point cloud.

The new gate:

- resolves the named TIAGo and `potted tree(2)` directly from the checked-in `office.wbt`, instead of copying simulator coordinates into runtime or verifier code;
- keeps the fixture dimensions and their pinned Cyberbotics R2025a source in a dedicated test metadata file;
- associates by metric position before checking the label, so a wrong class cannot evade measurement;
- samples the live Scene state eight times and reports target recall, label accuracy, stable-ID count, median center XY/Z error, maximum XY drift, yaw-oriented 3D bbox IoU, point-cloud inlier fraction and point count, and navigation-grade fraction;
- retains the existing occupancy-bound, bbox-sanity, DBSCAN, and background-label checks;
- documents that this is a Webots regression floor, not a substitute for physical RGB-D evaluation.

No runtime capability, robot geometry, frame, or environment constant is added by this PR. Simulator-specific truth remains under `testing/`.

## Why

Historical real run 30073019220 passed the former 17/17 suite while admitting a plant before pose/map readiness and allowing the same static object ID to drift 0.205 m in XY and 0.287 m in Z. `goal_near` and Navigation still passed, showing that contract-flow success alone was not an accuracy gate.

## Local validation

- `python3 -m unittest discover -s testing -p 'test_*.py' -v`: 14 passed, including a negative regression proving that wrong labels, unstable IDs, bad geometry, and missing cloud evidence fail the gate;
- `python3 testing/simulate_ci.py`: all 18 scenarios passed offline simulation (run in an isolated temporary venv with PyYAML because the macOS system Python lacks it);
- Black check, Python compile, commit-authorship check, public-file CJK scan, and `git diff --check`: passed;
- `scripts/check-docs.sh` could not run locally because macOS Bash 3 lacks `mapfile`; the equivalent changed-file language scan passed and GitHub Docs Check remains authoritative.

## Real Webots gate

A current-stack GPU-backed run is still required. The self-hosted runner currently has only about 758 MiB free while the workflow requires 4096 MiB, and existing workloads were not stopped. This PR must remain draft and unmerged until the real metrics are captured and any threshold/runtime regressions are resolved.

Stack: #196 → #197 → #198 → #199 → this PR.

Refs #177

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -25,10 +25,12 @@ job). You can also run it by hand against any booted deploy.
 | Scene perception | Real scene service over Webots RGB-D, camera intrinsics, and ConceptGraphs | Scenarios require non-robot objects from the live scene registry and use one as the navigation target. |
 | Speech/audio hardware | Real speech service plus audio driver using ALSA `null` devices | The runner has no physical speaker/mic. CI checks TTS-to-speaker and mic/speaker driver wiring through normal ALSA devices, not acoustic output. |
 
-This workflow is therefore not a perception-quality or physical-audio quality
-test. It is a build/boot/dispatch/runtime integration test for the
-Webots-backed robot stack, with deterministic substitution only at the VLM
-planning boundary.
+This workflow is primarily a build/boot/dispatch/runtime integration test for
+the Webots-backed robot stack, with deterministic substitution only at the VLM
+planning boundary. It also contains one bounded Scene perception-quality gate
+against checked-in Webots ground truth. That gate does not replace physical
+robot evaluation. Audio remains a wiring check rather than an acoustic-quality
+test.
 
 ## Pieces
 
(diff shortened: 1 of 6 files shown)


enkerewpo:
Consolidation note: GitHub automatically marked this PR as merged when its exact head commit was fast-forwarded into its base branch (#199). No merge was performed into dev or dev-next. The commit now lives unchanged in draft PR #199, which remains unmerged.

PR #203, opened 2026-07-28
fix(rbnx): own deployment package lifecycle

enkerewpo:
## What changed

- make `rbnx boot` the sole owner of primitive, service, and skill package processes
- start and drive primitive lifecycle before dependent system services
- start skills directly, wait for Atlas registration and `Driver(INIT)`, and print each final state
- reduce Soma to body description/runtime-state service only
- remove the private `ROBONIX_SOMA_STAGE_FD` pipe, deployment-manifest parsing, and duplicate package launcher from Soma

## Root cause

Soma and rbnx split deployment ownership across an internal two-stage protocol. rbnx only waited for primitive stage 1, then wrote the skill trigger and reported boot success without waiting for skill startup. This made skills absent from boot output and allowed a deployment to look ready before a skill had registered.

## Impact

Boot now has one process owner and one lifecycle path. A skill such as Explore is visible as `INACTIVE (skill — awaits executor activate)` only after its provider and MCP contracts are registered; failures are reported by `rbnx boot` instead of being hidden in `soma.log`.

## Validation

- `cargo test -p robonix-cli`: 27 passed
- `cargo test -p robonix-soma`: 17 passed
- focused Rust formatting check passed for modified files
- real WorkPC Webots boot: all 19 components started; five primitives reached ACTIVE; system services and navigation reached ACTIVE; Explore registered and reached INACTIVE; no Soma stage output remained
- real `rbnx shutdown`: all 19 owned components were stopped and the deployment state file was removed

Draft for maintainer review; do not merge yet.


enkerewpo:
Closed because this was an urgent WorkPC-local screenshot fix; no PR requested.

PR #204, opened 2026-07-28
fix(rbnx): report Soma-managed boot readiness

enkerewpo:
## Summary

- preserve Soma as the sole owner of Primitive and Skill process launch, lifecycle transitions, runtime records, and shutdown
- make the public boot order consistently `system -> primitive -> service -> skill`
- wait for every exact Primitive provider to become ACTIVE and every Skill provider to finish INIT before reporting boot success
- remove the internal stage-2 trigger acknowledgement from user-facing output
- clean root-owned Webots evidence through a bounded BusyBox fallback when the self-hosted runner has no passwordless sudo

## Why

The existing two-stage Soma architecture was already present on `dev`, but `rbnx boot` split System output around Primitive and returned success immediately after writing the Skill trigger. This made a healthy Explore Skill look absent. Separately, dev run 30333781611 failed before checkout because container-created root-owned logs could not be removed by the runner user.

## Verification

- `cargo test -p robonix-cli`: 25 passed
- `cargo clippy -p robonix-cli --all-targets -- -D warnings`: passed
- WorkPC Keystone integration: 6 focused deploy tests passed and `make install` completed
- real Webots boot: exact section order `system -> primitive -> service -> skill`
- five Primitive providers reached ACTIVE and were reported as Soma-managed
- Explore reached INACTIVE after INIT and exposed driver, start, status, and cancel capabilities before boot success
- process evidence showed Explore launched directly under Soma; rbnx state excluded Primitive and Skill children
- `rbnx shutdown` removed Soma, Explore, Atlas, and all booted providers without residue
- root-owned cleanup fallback removed both a controlled proof directory and the stale CI log directory

## dev-next audit

`dev-next` still keeps Primitive and Skill ownership in Soma and already waits for Skill readiness, but it currently prints `system -> primitive -> system -> service -> skill`. After this dev fix is reviewed and merged, the grouping correction should be synchronized into dev-next without pulling its experimental Pilot changes back into dev.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -388,32 +388,6 @@ jobs:
           # The capacity gate deliberately runs before checkout. Self-hosted
           # workspaces persist between jobs, so remove only generated evidence
           # from the previous run before a pre-checkout failure can upload it.
-          remove_generated_path() {
-            root="$1"
-            relative="$2"
-            target="$root/$relative"
-            [ -e "$target" ] || return 0
-            if rm -rf -- "$target" 2>/dev/null; then
-              return 0
-            fi
-            # Docker-backed providers can leave root-owned files in a
-            # bind-mounted log directory. This runner has no passwordless
-            # sudo, so use the same bounded container cleanup as the full
-            # workspace reset below.
-            if ! docker image inspect busybox:latest >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


enkerewpo:
Closing per maintainer request. The runner failure was WorkPC environment residue, not a workflow-code defect; no CI workflow change should be proposed here. The validated boot change remains on the branch for direct follow-up.

PR #205, opened 2026-08-02
feat(sentinel): add authenticated robot policy enforcement

enkerewpo:
## What changed

- add a typed Sentinel rule engine and `ListRules` / `ReplaceRules` management contracts
- carry a Keystone-validated session on the internal Liaison -> Pilot -> Executor path, redact it from public Plan events, and re-resolve identity before execution
- enforce robot-level rules immediately before capability dispatch using typed RFC 6901 state paths, user/role, capability, time, and region inputs
- require an authenticated Keystone admin for rule management
- advertise the Sentinel provider, validate explicit and derived listen-port collisions, and bootstrap a versioned Webots demo policy only on first run

## Why

This provides the minimum end-to-end Sentinel loop for the Keystone demo branch without hard-coding robot predicates such as `moving` or `gripper_open`. Robot state remains generic and typed, so deployments can address boolean, integer-range, and float-range conditions through stable state paths.

## Security properties

- caller-supplied identity is overwritten after Keystone resolution
- Executor derives canonical identity from Keystone instead of trusting Plan user/role fields
- the reusable session token is cleared from client-visible Plan events
- malformed capability arguments and missing/wrongly typed policy state fail closed
- only Keystone admins can list or replace rules

## Validation

- `cargo test -p robonix-sentinel -p robonix-executor -p robonix-cli` (7 Sentinel, 44 Executor, 32 CLI tests)
- `cargo clippy -p robonix-sentinel -p robonix-executor -p robonix-cli --all-targets -- -D warnings`
- `python3 -m unittest scripts/tests/test_check_commit_authorship.py`
- `python3 scripts/check_commit_authorship.py --base origin/dev-keystone --head HEAD`
- `git diff --check`

## Companion change

The admin rule-management page is being published separately from the robonix-client Keystone branch.

This PR intentionally targets `dev-keystone`; it does not change `dev` or `dev-next`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3217,6 +3217,7 @@ name = "robonix-executor"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "chrono",
  "clap",
  "nix 0.31.3",
  "prost",
@@ -3225,6 +3226,7 @@ dependencies = [
  "robonix-atlas",
  "robonix-codegen",
  "robonix-scribe",
+ "robonix-sentinel",
  "serde",
  "serde_json",
  "serde_yaml",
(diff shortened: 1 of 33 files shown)


enkerewpo:
Companion Client PR: syswonder/robonix-client#9 (targets `feat/keystone-accounts`). Please review/land the backend contract before or together with the Client UI.

PR #220, opened 2026-08-06
fix(pilot): restore closed-loop replan boundary; make task_update goal echo optional

enkerewpo:
## What

`task_update` required `goal`, `success_criterion`, and `status` together. `apply_task_update` then **rejected** any goal that differed from the harness copy — so the echoed goal was never applied. It only ever served as a staleness token: one bit of information, at a cost of `O(len(goal))` output tokens on **every** planning round.

That is invisible for a short goal and dominant for a long one.

## Evidence

Measured on an EB-Habitat episode whose task text lists a 70-entry public action catalog. Across the 7 planning calls of a single successful episode:

| part of generated output | chars | share |
|---|---|---|
| `task_update.goal` echo | 20364 | **86%** |
| the RTDL plan itself | 1306 | 6% |
| content / descriptions | 1996 | 8% |

Per-call latency was **27.3s ±4.6**, against **13–17s** for callers with smaller payloads on the same model and the same endpoint, measured minutes apart in both directions — so this is workload, not server variance.

## Change

- `goal` and `success_criterion` become optional; omitting `goal` keeps the current one.
- Completion still requires the verbatim echo: `status: "done"` is honoured **only** when the model reproduced the goal. A reply sampled before a newer steer therefore still cannot silently finish the task. Every other transition is recoverable, so ordinary rounds can send just `{"status":"in_progress"}`.
- The prompt contract tells the model to omit `goal` on ordinary rounds and include it only when finishing.
- Unknown keys are now rejected explicitly (previously caught by the exact-arity check).

Sending `goal` every round remains valid, so existing model behaviour still parses — this is backward compatible.

## Tests

`cargo test -p robonix-pilot` — 57 passed. Five new cases cover: `status` required, unknown key rejected, `goal`/`success_criterion` optional, omitted goal keeps the harness goal and still progresses, `done` without echo stays in progress, `done` with matching echo finishes.

Found while building the EmbodiedBench V13 comparison harness.

### DIFF
--- a/system/pilot/src/planner.rs
+++ b/system/pilot/src/planner.rs
@@ -87,6 +87,26 @@ pub(crate) struct TaskState {
     status: String,
 }
 
+/// A model-supplied change to the standing task.
+///
+/// `goal` and `success_criterion` are optional so an ordinary round does not
+/// have to reproduce the whole goal verbatim. The harness owns the goal text;
+/// the echo only ever served as a staleness token, so its information content
+/// is one bit while its cost is O(len(goal)) output tokens on EVERY round. On
+/// a long goal — an embodied benchmark that lists a 70-entry action catalog,
+/// say — that echo dominated generation and inflated per-call latency.
+///
+/// Omitting `goal` therefore keeps the current one. To stay safe, completion
+/// still requires the verbatim echo: `status: "done"` is honoured only when
+/// the model reproduced the goal, so a reply sampled before a newer steer can
+/// never silently finish the task. Everything else is recoverable.
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct TaskUpdate {
+    goal: Option<String>,
+    success_criterion: Option<String>,
+    status: String,
+}
+
 const DEFAULT_SUCCESS_CRITERION: &str =
     "The user's request is completed and the result has been verified.";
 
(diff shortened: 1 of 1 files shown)


enkerewpo:
## Update: this PR now carries both Pilot fixes found by the same study

The second commit restores `ROBONIX_PILOT_REPLAN_AFTER_EACH_CALL_CONTRACTS`, which the current baseline stopped reading while callers still set it.

**Evidence** — EB-Habitat `base/2`, same model and cell for all three systems:

| system | planning turns | actions dispatched | invalid |
|---|---|---|---|
| Robonix (this baseline) | **1** | **28** | **12** |
| a ReAct baseline (replans每step) | 22 | 22 | 9 |
| a baseline replanning each round | 4 | 4 | 0 |

One turn dispatched 28 actions with nothing reconsidering the plan after the first failure. The harness had set the variable (`run_habitat_episode.py:379`) since V12; it was silently ignored.

It also flatters easy cells: on `base/0` the same behaviour gave 1 turn / 4 actions / 0 invalid, which reads as a multi-step planning win when half of it is an absent feedback loop.

Both commits are independent and each has its own tests. 60 Pilot tests pass, `cargo fmt --check` and `clippy -D warnings` clean.

PR #228, opened 2026-08-09
merge: fold dev-next into dev

enkerewpo:
Folds `dev-next` into `dev`. `dev-next` is **not** modified — it stays as the release/demo branch for the upcoming on-site work.

## Why this was not a fast-forward

The branches had diverged in both directions: 114 commits on `dev-next`, 94 on `dev`, common ancestor `26db769a`. 62 files conflicted. 23 commit pairs shared a subject across both branches, 14 of them byte-identical patches — the same fix applied twice.

## How each conflict was decided

Not "take one branch". Per subsystem, on evidence:

| Area | Taken from | Reason |
| --- | --- | --- |
| Lifecycle drivers — `rbnx` launch/deploy, `soma` | dev-next | The target architecture. It adds `registration_id` to the Atlas provider. |
| dev's fingerprint takeover detection (469 lines) | dropped | It existed only because `registration_id` did not. `is_new_provider_registration` reads one field instead of opening a metadata-only `ConnectCapability` probe. |
| Pilot | dev | Three weeks newer, and carries the compact planning context from #210. Ten of dev-next's eleven pilot commits already had an equivalent on dev. |
| `camera/extrinsics` + Scene transform code | dev-next | TF tree authoritative, contract as compatibility fallback — matches the vendor-onboarding guide. dev's README still claimed Scene never uses tf2, which the code had already outgrown. |
| `soma/msg/SensorExtrinsic.msg` | deleted | That migration finished on dev-next. Nothing references it. |
| `memory/srv/*` | dev | ScribeMem deprecation notes postdate dev-next's copies. |
| Vitals bind address | dev (`127.0.0.1`) | Follows the provider-loopback direction of #188, not dev-next's `0.0.0.0`. |
| `GetUrdf` assets | contract from dev, implementation ported | The contract kept `include_assets`/`assets`, so dev's asset reader was ported onto dev-next's service. |
| Codegen interpreter pinning | dev | `RBNX_CODEGEN_PYTHON` is used by 10 scripts on dev versus 5 for dev-next's wrapper, and postdates it. |
| dev-next's three CI workflows | dropped | They existed to run checks on `dev-next` itself. |
| Webots manifest, `sim/start.sh` | union | Each branch added an independent feature. |

## Validation

- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo test --workspace --all-targets` — 296 passed, 0 failed
- Every symbol introduced by dev's 70 unique commits still resolves in the merged tree. Two flagged by that sweep were checked by hand: the fingerprint machinery (deliberately dropped, above) and `cancel_all_executor_plans` (never existed on dev — the cancel-all path is dev's `MetaPlanOp::CancelAll`, which is present).

Four merge defects were found and fixed rather than shipped: a duplicated `config_json` clap argument, four duplicated helpers in `soma/store.rs`, a duplicated executor test, and `is_builtin_system` being called without a definition after taking dev-next's `deploy.rs`.

## Not included

- Three dev-next Scene features (semantic map persistence, epoch-aware annotations, map identity discovery) overlap #199's files and are left for after it lands.
- dev-next's per-instance simulator teardown: `runtime_state.sh` is merged and used by `start.sh`, but splicing it into `stop.sh` produced invalid shell twice, so `stop.sh` is dev's. Worth a follow-up.

## Update 2026-08-21

- Caught the branch up with both sides: merged the 10 commits dev gained since (CI pins, README/contributors cleanup, Assisted-by checker) and dev-next's `f6cacfc4` (URDF out of Pilot's prompt; one trivial conflict — dev-next's older file predates dev's `compact_yaml` tests, tests kept).
- The "Not included" note below is now stale: after #199 landed, the three dev-next Scene features (semantic map persistence, epoch-aware annotations, map identity discovery) reached dev through their own PRs and are present in this branch.
- `fix(liaison)`: the runner's newer clippy flags `chunks_exact(2)`; switched to `as_chunks::<2>()`.
- Full Webots e2e re-dispatched via workflow_dispatch (run 32470673781); the earlier "no scenario summary" result came from a runner-side infrastructure issue, since resolved.


enkerewpo:
@robonix-ci test

PR #230, opened 2026-08-14
fix(speech): split long Tencent TTS input

### LINKED ISSUE #211: [speech][Tencent TTS] Long responses fail with UnsupportedOperation.TextTooLong
## Summary

Long Robonix responses are passed to Tencent Cloud TTS as a single `TextToVoice` request. Tencent rejects sufficiently long input with `UnsupportedOperation.TextTooLong`, so the text response is displayed but speech is skipped.

This was reproduced from Robonix Client with a 244-character Chinese response:

```text
SYNTHESISING 244 CHARS
tts skipped: tts error: Tencent TTS UnsupportedOperation.TextTooLong: Text too long
```

Short progress utterances (for example, 11–17 characters) synthesize and play normally.

## Root cause

On `dev-next`:

- `services/speech/speech_service/tencent_cloud.py`: `TencentTTSBackend.synthesize()` places the complete input in `payload["Text"]` and sends one request.
- `TencentTTSBackend.synthesize_stream()` calls `synthesize()` first and only chunks the returned PCM bytes, so it does not avoid Tencent's input limit.
- `services/speech/speech_service/service.py`: `speech/speak` also calls `tts_backend.synthesize(text)` once.

The Client is only surfacing the server-side TTS error.

## Expected behavior

Long responses should be spoken completely without requiring callers or clients to know provider-specific text limits.

## Suggested fix

- Split long text at sentence punctuation first (`。！？；\n`), then commas or a safe hard boundary when necessary.
- Keep each Tencent request below a configurable provider-safe limit.
- Preserve segment order.
- For one-shot TTS, concatenate compatible PCM segments before returning.
- For streaming TTS, synthesize/yield segments sequentially instead of synthesizing the whole input first.
- Keep the existing behavior for short text.
- Report which segment failed while retaining the original exception context.

## Acceptance criteria

- A Chinese response longer than the Tencent single-request limit plays fully.
- Mixed Chinese/English text is segmented safely.
- Text at and around the configured boundary is covered by tests.
- `tts`, `tts_stream`, and `speech/speak` do not regress for short input.
- No `UnsupportedOperation.TextTooLong` occurs for inputs handled by the service splitter.

i6bimua:
## Problem

Tencent TextToVoice rejects long Chinese input with
`UnsupportedOperation.TextTooLong`. This prevents longer utterances from being
returned or played completely through unary TTS, streaming TTS, and
`speech/speak`.

## Changes

- Split Tencent TTS input into provider-safe segments.
- Prefer sentence punctuation, then clause or whitespace boundaries, and
  finally bounded hard splits.
- Preserve every accepted input character and synthesize segments in order.
- Keep short-text behavior as a single provider request.
- Use one primary language selection for the complete utterance.
- Add total-character, request-count, decoded-PCM, and cache limits.
- Stop requesting later segments after unary or streaming cancellation.
- Validate PCM/Base64 responses and retain Tencent request IDs in errors.
- Preserve unary response error semantics and explicit streaming gRPC errors.
- Document Tencent TTS configuration and limits.

## Compatibility

- No capability contract or generated-code changes.
- No migration is required.
- Tencent output remains 16 kHz mono `pcm_s16le`.
- `TENCENT_TTS_CODEC` must remain `pcm`.
- Added configurable per-request and total-text limits.

## Validation

- Speech service tests: 47 passed.
- Tencent TTS tests on Python 3.10: 27 passed.
- Tencent TTS tests on Python 3.11: 27 passed.
- `cargo clippy --workspace --tests -- -D warnings`
- `cargo build --workspace`
- `cargo test --workspace --all-targets`: 251 passed.
- Runtime portability, Webots smoke, LaserScan regression, documentation, Ruff,
  and commit-authorship checks passed.

No screenshots are applicable because this is a backend-only change.

Closes #211

### DIFF
--- a/services/speech/speech_service/tencent_cloud.py
+++ b/services/speech/speech_service/tencent_cloud.py
@@ -16,6 +16,7 @@ import logging
 import os
 import random
 import time
+import unicodedata
 import uuid
 from dataclasses import dataclass
 from typing import Iterator
@@ -24,7 +25,27 @@ from urllib.parse import quote, urlencode
 import requests
 from websockets.sync.client import connect
 
+from speech_service.tts_errors import TTSInputError
+from speech_service.tts_runtime import tts_request_is_active
+
 log = logging.getLogger(__name__)
+
+_TTS_DEFAULT_MAX_CHARS = 140
+_TTS_PROVIDER_MAX_CHARS = 150
+_TTS_DEFAULT_MAX_TOTAL_CHARS = 5_000
+_TTS_MAX_SEGMENTS = 40
+_TTS_MAX_TOTAL_CHARS = _TTS_PROVIDER_MAX_CHARS * _TTS_MAX_SEGMENTS
+_TTS_MAX_ONE_SHOT_PCM_BYTES = 4 * 1024 * 1024 - 64 * 1024
+_TTS_MAX_SEGMENT_PCM_BYTES = _TTS_MAX_ONE_SHOT_PCM_BYTES
+_TTS_MAX_TOTAL_PCM_BYTES = _TTS_MAX_ONE_SHOT_PCM_BYTES
+_TTS_SENTENCE_BREAKS = frozenset(
+    "\N{IDEOGRAPHIC FULL STOP}"
+    "\N{FULLWIDTH EXCLAMATION MARK}"
+    "\N{FULLWIDTH QUESTION MARK}"
+    "\N{FULLWIDTH SEMICOLON}"
+    ".!?;\n"
+)
+_TTS_CLAUSE_BREAKS = frozenset("\N{FULLWIDTH COMMA}, \t")
 
 
 def _env_first(*names: str, default: str = "") -> str:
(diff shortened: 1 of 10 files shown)


enkerewpo:
@robonix-ci test

services/speech/speech_service/tencent_cloud.py
```diff
@@ -43,6 +64,56 @@ def _require_env(*names: str) -> str:
...
+    if max_chars <= 0:
+        raise ValueError("Tencent TTS max chars must be greater than zero")
+    if len(text) <= max_chars:
+        return [text]
+
+    segments: list[str] = []
+    remaining = text
+    while len(remaining) > max_chars:
+        window = remaining[:max_chars]
+        split_at = (
```
kaileliu, line 130: **Blocking:** choosing any sentence boundary in the current window can create tiny follow-up segments and make otherwise valid input exceed `_TTS_MAX_SEGMENTS`. With the defaults, `text = ("a" * 141 + ".") * 35` is 4,970 characters (below the 5,000-character limit), but this splitter produces 70 alternating 140/2-character segments, so `_segments_for_provider` rejects it. A hard bounded split needs only 36 requests and fits every configured limit. Please rebalance/ignore boundaries that create inefficient fragments, or fall back to a bounded split when the preferred result exceeds the request budget, and add this interaction as a regression test.

services/speech/speech_service/tencent_cloud.py
```diff
@@ -264,52 +335,205 @@ class TencentTTSBackend:
...
+        if len(segments) > 1:
+            log.info(
+                "Tencent TTS split %d chars into %d requests",
+                len(text),
+                len(segments),
+            )
+        audio = []
+        audio_size = 0
+        for index, segment in enumerate(segments, start=1):
+            if not tts_request_is_active():
```
kaileliu: **Blocking:** this cancellation check is never connected to the `speech/speak` path. FastMCP's async shim invokes the synchronous `speak()` inline, and `speak()` blocks that event-loop thread on `ThreadPoolExecutor(...).result()` without binding `bind_tts_request_active`; therefore this function sees the ContextVar default and every later segment still runs after the MCP request is cancelled/disconnected. A default-valid 5,000-character input can fan out to about 36 paid HTTP calls (up to 40 by policy), blocking all other MCP work for the full sequence and still driving the speaker afterward. Please make the MCP path non-blocking/cancellable and propagate cancellation so no later segment is requested; cover cancellation after the first segment.

kaileliu (dismissed):
The Tencent request validation, PCM checks, and focused tests are substantial, and the existing checks are green. Two lifecycle/segmentation issues still block the main long-text path: valid bounded input can be rejected because natural-boundary splitting unnecessarily exceeds the request cap, and `speech/speak` cannot observe cancellation while its expanded provider fanout blocks FastMCP's event loop. Please address the two inline findings and add regression coverage for both interactions.

i6bimua:
Addressed both blocking findings:

1. Segmentation: ignore inefficient early punctuation boundaries that
   inflate segment counts, and fall back to a hard budget split when the
   natural result would exceed `_TTS_MAX_SEGMENTS`. The reviewer fixture
   `("a" * 141 + ".") * 35` now stays within the request cap.
2. `speech/speak`: made the MCP handler awaitable (no event-loop
   `.result()` block), bound `tts_request_is_active` for the synthesis
   worker, and stop later provider segments / playback after cancel.

Added regression coverage for both interactions. Ready for re-review.

i6bimua:
@robonix-ci test

enkerewpo:
@robonix-ci test

enkerewpo (approved):
LGTM, thanks

enkerewpo:
## Test results

Ran against PR head `e40be36` with real Tencent Cloud credentials.

`tests/test_tencent_cloud.py`: 31/31 pass.

| Case | Result |
|---|---|
| 368-char input | Split into 3 segments (124/128/116); no character lost, none over the 140 limit; 2155484 bytes PCM ≈ 67.4 s @16 kHz |
| 14-char input | Still one segment, one request |
| Cancellation | `cancelled before segment 2/3`; remaining segments never sent |

PR #232, opened 2026-08-15
feat(liaison): let a client end voice capture and submit early

enkerewpo:
A voice turn ends when the ASR backend reports end-of-utterance, which it derives from silence. In a room with continuous background noise that condition may never hold — chunks stay above `VAD_SPEECH_RMS`, the pump keeps recording, and the speaker cannot hand the turn over until `record_seconds` expires. The only escape today is dropping the `StartVoiceSession` stream, which discards everything already said. This came up on the Ranger Mini during a live demo.

This adds `robonix/system/liaison/voice/finish` as the manual equivalent of that automatic endpoint: capture stops, the ASR request sender is dropped so the backend flushes its final transcript, and the turn proceeds down the normal submission path. The audio already spoken is kept, not thrown away. **This is deliberately not an abort** — dropping the stream remains the way to discard a turn.

## How it works

Running captures register an `AtomicBool` in a process-wide map keyed by session id. `finish_capture` flips it and the mic pump observes it on its next chunk, breaking the loop exactly the way the silence path does. The pump previously blocked on `mic_stream.message()` until the record deadline, so it now polls at `FINISH_POLL_INTERVAL` to notice the flag while audio is still flowing. A drop guard deregisters the session however the pump returns, so a finish request for a turn that already ended reports `ok=false` rather than claiming success against a dead session.

Registering the contract with Atlas is what lets clients tell an upgraded Liaison from an older one: a Liaison without this build never declares it, so a client hides its finish control instead of failing with `UNIMPLEMENTED`.

## Client side

[robonix-client 1.0.0](https://github.com/syswonder/robonix-client/releases/tag/v1.0.0) already ships the caller (`Stop recording`, bound to `F2`). It probes Atlas for this contract every 7s and keeps the control hidden until a robot registers it, so the client works unchanged against Robonix 1.0.0 — the button simply never appears. Merging this is what turns it on.

## Validation

- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- Exercised end-to-end on the AgileX Ranger Mini v3 deployment before the 1.0 branch reset

Rebased onto current `dev`. The original commit was written against `dev-next`, whose liaison carries the lifecycle-driver restructure; the additions were re-applied by hand onto `dev`'s own server-builder and capability-declaration blocks rather than dragging that scaffolding along.


system/liaison/src/voice.rs
```diff
@@ -660,13 +713,27 @@ async fn stream_capture_and_recognize(
     let detected_rate_pump = Arc::clone(&detected_rate);
     let _language_owned = language.to_string();
 
+    // Publish this capture so `robonix/system/liaison/voice/finish` can end
+    // it early; the guard deregisters it however this function returns.
+    let finish_flag = Arc::new(AtomicBool::new(false));
```
kaileliu, line 718: **Blocking:** registration happens after `KIND_RECORDING_STARTED` is sent and after `mic_client.mic(...).await`. The released client exposes the Stop recording control as soon as it receives that event, so an immediate click calls `finish_capture` before this entry exists, gets `ok=false`, and then recording continues once the mic RPC completes. Register/preserve a pending finish before publishing the recording-started state (or publish that event only after registration), and cover the delayed-mic startup race with a test.

system/liaison/src/voice.rs
```diff
@@ -660,13 +713,27 @@ async fn stream_capture_and_recognize(
...
+    {
+        let mut guard = match ACTIVE_CAPTURES.lock() {
+            Ok(guard) => guard,
+            Err(poisoned) => poisoned.into_inner(),
+        };
+        guard.insert(session_id.to_string(), Arc::clone(&finish_flag));
```
kaileliu, line 724: **Blocking:** `session_id` is caller-supplied, so this unconditional insert can replace another live capture. A finish then targets only the newer flag, and when either `CaptureRegistration` drops its key-only `remove` can delete the other capture's registration. Please atomically reject duplicate active IDs or make registration/removal identity-aware, with a concurrent duplicate-ID regression test.

system/liaison/src/voice.rs
```diff
@@ -660,13 +713,27 @@ async fn stream_capture_and_recognize(
...
+    }
+    let _capture_registration = CaptureRegistration(session_id.to_string());
```
kaileliu, line 726: **Blocking:** this guard is owned by `stream_capture_and_recognize`, not by the mic pump. After VAD/deadline/EOF stops the pump, the function can still wait indefinitely for the ASR final event, while `finish_capture` returns `true` and claims a capture is finishing even though no pump consumes the flag. Tie deregistration to actual pump termination (including cancellation/error paths) and test cleanup while ASR final delivery is delayed.

kaileliu (changes requested):
The capability and client integration are well motivated, and the existing CI is green. However, the capture registry currently has lifecycle races that can make the new stop action report the wrong result or lose control of a live capture. Please address the three inline findings and add focused regression tests for immediate finish during startup, duplicate active session IDs, and registry cleanup when the mic pump exits.

kaileliu (approved):
Approving the capability and client-facing behavior. The three previously raised capture-registry race cases remain valid follow-up work, but are treated as non-blocking for this merge direction: immediate finish during mic startup, duplicate active session IDs, and deregistration at actual pump termination.

PR #234, opened 2026-08-22
feat(scene): correctable, disappearing objects on upstream ConceptGraphs

### LINKED ISSUE #177: [BUG] Scene object world model retains inaccurate, stale, and off-map ghost objects

enkerewpo:
Rebuild of #199 on top of `dev`, per its review. **Upstream ConceptGraphs stays the perception core** — every feature here hooks in from outside the `MapObjectList`; the twelve upstream symbols Scene uses are untouched.

Closes #177. Supersedes #199 (draft), which reached the right contracts and the right absence-detection algorithm but replaced the ConceptGraphs SLAM kernels with an in-repo reimplementation, undisclosed, inside a 13.5k-line commit. This PR is `+1,843 −7` over 20 files against #199's `+15,792 −2,089` over 69.

## What it does

- **Correction contracts** (`list/relabel/geometry/delete/flush`, ported unchanged from #199's review): every call asserts the map epoch it was read under (`ListObjects` now echoes `map_id` + `generation`); an edit aimed at a stale map is rejected, not applied.
- **Registry-level operator state**: sticky verbatim labels with explicit undo, provenance-marked non-navigation-grade geometry overrides, robot self-record untouchable. Operator-corrected objects are immune to soft mark-missing and TTL deletion — a relocated object's stale cloud must not erase the human's edit.
- **Epoch-checked mutation coordinator**: single writer for derived-object mutations, snapshot persistence with rollback; deletion drops the bound CG entries under the inference lock so the next tick cannot resurrect the record.
- **Visible-absence eviction** (#177's actual ask): unmatched objects' stored clouds are projected into the current depth image; a miss counts only where the measured surface is materially *behind* the stored object (closer = occlusion, similar = detector miss, invalid = unknown — none delete state). Removal needs `visibility_miss_ticks` consecutive clear verdicts, then soft-evict → TTL, reversible by re-observation until the prune.

## Review findings from #199, resolved here

| Finding | Resolution |
| --- | --- |
| ConceptGraphs silently replaced by 760-line in-repo kernels | Not ported. Upstream imports (`slam_classes`, `mapping`, `slam.utils`, `model_utils`) unchanged; the aarch64 concern was moot — `open3d>=0.17,<0.19` ships aarch64 wheels, which is why that pin exists |
| Operator corrections deletable via visible-miss → TTL | Immune at the registry level (both mark-missing and prune guard on operator attributes) |
| One label edit wiped the whole caption/relation cache | `invalidate_object` drops only that object's entries; the full clear is reserved for flush |
| Operator labels silently canonicalized (desk → table) | Stored verbatim |
| Vocabulary config could silently disable perception | N/A — the baked-checkpoint/TensorRT machinery this applied to was not ported; model loading is unchanged `dev` |

## Deliberately not ported

Web correction UI (+542), TensorRT cache (+422), model storage/baking (+210), lifecycle runtime (+240), perception profiles, and #199's 34-file test suite — the ten tests here cover the decisions (epoch rejection, guards, verbatim labels, narrow invalidation, absence verdicts, streak gating), not the plumbing. Each can be its own reviewable PR if wanted.

## Validation

- `system/scene` pytest: 146 passed, 30 skipped locally; the 10 failures are identical on `dev` (missing codegen env on this machine) — full suite + Webots runs on the GPU runner next.
- Scene benchmark harness (`scene-benchmark-harness` branch) scores via Scene's HTTP surface, so the five-world numbers can be re-measured on this tree as-is.


enkerewpo:
Reviewed at `d51ec610`. The design holds up: epoch assertions on every mutation, rollback of runtime state when the snapshot write fails, verbatim operator labels, robot-self and operator-correction immunity carried consistently through `soft_evict` and `prune_expired`, and a visible-absence rule that puts the burden of proof on deletion (occlusion, detector miss, and invalid depth all decline to delete). The three new test files cover the paths that matter. All checks are green.

Two notes, neither blocking.

**1. A zero-detection tick skips absence evaluation entirely.** In `_tick_locked`, the `len(det_list) == 0` branch returns after `_maybe_periodic_cleanup()`, so neither `_update_visible_absence` nor `_project_to_registry` runs. That is precisely the case the feature targets: the robot faces a spot where a mapped object used to be, the frame yields no detections at all, and the map keeps the object. With a 55-term closed vocabulary, empty ticks are not rare. Moving the absence update (and the registry projection) above that early return would close the gap; the streak counter already protects against a single noisy frame.

**2. `system/liaison/src/voice.rs` does not belong in this PR.** The change rewrites `chunks_exact(2)` as `as_chunks::<2>()` in `pcm_rms_s16le`. It is behaviourally identical, unrelated to Scene, and raises the effective minimum Rust version for a repository that pins no toolchain. Suggest dropping it here and landing it separately if it is wanted at all.

Neither needs to hold the merge. Filing (1) as a follow-up is fine.

enkerewpo:
Field-design audit of the contract surface, at `54f465ff`.

The five `.toml` contracts match the shape of `list_regions` and `get_scene_graph` exactly: `id` / `version` / `kind = "service"` / `idl` / `description`, plus `[mode] type = "rpc"`. Nothing to change there.

On the IDL, three things needed correcting and are now fixed:

**`observation_count` carried two types.** It was added to `Object` as `uint64`, but `SceneGraphNode` has published the same quantity under the same name as `int32` since before this branch. One name and one meaning must not carry two types across a single IDL library, and `SceneGraphNode` is already on the wire, so `Object` follows it. Unsigned is arguably the better model for a count, but changing the published field is a breaking change and does not belong in this PR.

**`note` truncated silently.** The coordinator strips whitespace and caps the stored text at 512 characters; the contract said nothing, so a longer note was quietly cut. All four services that accept a `note` now state the behaviour.

**`Object.frame_id`'s comment was wrong for one producer.** It said an empty value means the entry has no metric frame. `GetObjectContext.nearby_objects` is built from scene-graph nodes, whose coordinates *are* map-frame but which carry no frame string. The comment now says what empty actually means and what follows from it: `UpdateObjectGeometry` rejects an edit built from such an entry rather than applying it blind.

One inconsistency is noted but deliberately left alone: counts across this library are mixed — `SceneGraphNode.observation_count` and now `Object.observation_count` are `int32`, while `FlushObjects.deleted_count` is `uint64`. Unifying them is a separate change against the published surface, not something to fold in here.

Scene suite: 202 passed. The one failure, `test_runtime_protobuf_contract`, needs Docker and fails identically on the unmodified base.

PR #239, opened 2026-09-01
fix(ci): stop the Webots suite failing on a flaky route to GitHub

enkerewpo:
The Webots integration suite has been failing on `dev` because git could not reach GitHub, not because anything in the deployment was wrong. Two distinct causes, both fixed here, plus a third failure that surfaced once the network ones were gone.

**A dead proxy the route decision could not see.** The runner service inherited `HTTP_PROXY=127.0.0.1:9981` from a macOS-forwarding config that was disabled on 2026-08-10. The listener is still up but forwards nothing, so clones died with `GnuTLS, handshake failed` — an error that names TLS and hides the fact that the proxy is a dead end. The mirror step now clears the inherited values for the job, probes each candidate route for real with `git ls-remote`, and when no route answers leaves git alone to go direct rather than pointing it at a closed port. Clash on this runner drops out for stretches at a time and is the only way out when ghfast.top is blocked, so its probe retries six times at ten-second intervals instead of treating one failure as absence.

**A route decided once, then trusted for twenty minutes.** Even a correct decision at job start does not survive Clash dropping midway through `rbnx build`, which clones several remotes over several minutes. Both clone sites gave up after a single attempt, so one handshake failure partway through the fetch phase failed a run that would have succeeded seconds later — this is what killed run 33527386073. Clones now go through a shared helper that retries three times with a 5s backoff, removing any partial checkout between attempts, since git refuses to clone into a non-empty directory and a half-written tree would otherwise turn one flaky moment into a permanent failure.

**A skill call racing its own CMD_INIT after the lifecycle reboot.** Run 33506916121 (dev at 750cc58f) got past the network and failed only the post-restart pass of `object_navigation`: `skill explore is REGISTERED; automatic activation is only valid from INACTIVE`. The atlas log shows the window: the explore container declared its capabilities at 11:41:24.5, the restart suite dispatched its first explore call at 11:41:25.48, and rbnx boot sent CMD_INIT (REGISTERED → INACTIVE) at 11:41:25.74. Both readiness checks let the suite start inside those 260 ms: the lifecycle validator only required the explore contracts to be declared and the service providers to be ACTIVE, and the workflow's post-restart gate only looked at nav2. Both waits now treat a skill as booted only once it has left REGISTERED (INACTIVE or ACTIVE). The executor deliberately never retries activation, so the fix belongs in the readiness checks, not in the executor.

## Validation

- `cargo fmt --all -- --check` and `cargo clippy -p robonix-cli --all-targets -- -D warnings` clean.
- `cargo test -p robonix-cli git_clone_with_retry` — new test seeds a partial checkout, clones a nonexistent repo, and asserts all three attempts ran and the partial tree was removed.
- The retry helper's behaviour was confirmed from the test output: three `fatal: repository ... does not exist` lines for one call.

The clone retry is a behaviour change for every `rbnx build`, not just CI: a genuinely bad URL now costs 10s of backoff before failing. That trade buys not losing a twenty-minute deployment build to one dropped packet.
- Webots suite dispatched on this branch after the readiness fix: https://github.com/syswonder/robonix/actions/runs/33770928429 (result to be recorded when it finishes).

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -530,6 +530,17 @@ jobs:
           # developer workstation, so CI must not mutate ~/.gitconfig.
           git_config="$RUNNER_TEMP/robonix-ci-gitconfig"
           : > "$git_config"
+          # The runner service inherited HTTP_PROXY=127.0.0.1:9981 from a
+          # macOS-forwarding config that was disabled on 2026-08-10. The
+          # listener is still up but forwards nothing, so every clone since has
+          # died with `GnuTLS, handshake failed` — an error that names TLS and
+          # hides the fact that the proxy is a dead end. Drop the inherited
+          # values for this job so the routing decided below is the one that
+          # applies.
+          for stale in HTTP_PROXY HTTPS_PROXY ALL_PROXY http_proxy https_proxy all_proxy; do
+            printf '%s=\n' "$stale" >> "$GITHUB_ENV"
+          done
+          unset HTTP_PROXY HTTPS_PROXY ALL_PROXY http_proxy https_proxy all_proxy
           if timeout 15s git ls-remote --exit-code \
               "https://ghfast.top/https://github.com/syswonder/robonix.git" \
               HEAD >/dev/null 2>&1; then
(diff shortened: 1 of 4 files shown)


enkerewpo:
Verified end to end. Webots Integration Tests pass on a branch carrying these commits plus the depth optical frame fix — run [33556140209](https://github.com/syswonder/robonix/actions/runs/33556140209), every step green including `Build Webots deployment`, the step that has been failing on dev.

Note for reviewers: the two earlier runs on this branch (33533335665, 33545907733) predate the clone-retry commit and should not be read as evidence either way.

PR #243, opened 2026-09-04
feat(scene): add DualMap as a selectable perception backend

enkerewpo:
Scene's object accuracy had stopped improving, and the honest reason was that its perception was ours: a hand-rolled association loop we kept tuning, with no external reference to say what level it was at. This adds a reproducible published system as a selectable backend, and — for the first time — runs the two side by side on the same world, map, tour and scorer. The result is not the one this PR set out to show: on object F1 the two are level once the map is right, DualMap is clearly better at naming things, and the largest single gain came from the vocabulary rather than from the backend.

## What changed

**`perception.backend: concept_graphs | dualmap`.** DualMap (Eku127/DualMap, RA-L 2025, Apache-2.0, pinned at 157235e) runs as a Scene detector: YOLO-World + MobileSAM/FastSAM + MobileCLIP-S2, its own keyframe rule, its own local map and its own self-merge. Nothing above the detector changes — the object registry, the geometric relation loop, the capability surface and the web views are the same. `concept_graphs` stays the default.

**The DualMap object lifecycle is a deployment setting, and this was a real defect.** DualMap's lifecycle is sized for a dataset replay that maps every frame: a track needs `stable_num` (8) observations to become stable, and one that leaves the `active_window_size` (10) most recent frames without getting there is deleted after `max_pending_count` rounds. A robot mapping keyframes at walking pace sees each object a handful of times, so the map emptied out behind it — one 180 s Webots office run went from 65 tracks to 1, and Scene ended with a single object. Recall went from 0.04 to 0.75 once these matched the observation rate.

**The vocabulary is a deployment setting too, and it is the largest lever in this PR.** DualMap ships a 101-name list built for homes — bed, bathtub, ironing board. A detector can only answer with a name it was given, so in an office it called the cabinet a sink and the chair an ironing board: of the 38 objects the benchmark counted as invented, 27 were within a metre of a real one. Right place, wrong name. `perception.dualmap.classes` takes the deployment's own list; the office manifests carry a 40-name office vocabulary, deliberately wider than what the world contains so the score still measures recognition rather than a lookup. Label accuracy went from 0.43 to 0.74 on that change alone.

**Labels are judged for meaning, not spelled the same.** The scorer matches by geometry and then compares labels; string equality called a monitor labelled `tv`, a couch labelled `sofa` and a plant labelled `vase` wrong, which is a statement about vocabulary rather than recognition. `testing/judge_scene_labels.py` asks a model whether each (expected, observed) pair names the same kind of object and writes the judgments file `evaluate_webots_scene.py` already accepts. It is strict on purpose — cabinet/shelf, keyboard/monitor and window/mirror are distinct — and on the office run it found exactly one equivalence (couch/sofa): with the right vocabulary, the remaining label errors are genuine confusions.

**Two merges that were tried and taken back out.** Both were written to attack duplicate objects, both reduced duplicates, and both were removed because they traded them for something worse:

- *Export-time filters* (containment across classes, same-class smear clustering, floor and size gates). On Replica they deleted every rug (IoU 0.16 → 0.00); the paper's own evaluation runs local-only, and reimplementing what the backend already does was the wrong place to fix it.
- *DualMap's own cross-view merge* (the abstract map, `global_map: true`) and a wider association voxel. Duplicates fell (31 → 6) while true positives fell with them (40 → 24) and ghosts rose — distinct nearby objects were being folded into one. Both stay available as opt-in settings; neither is on.

**mcp<2.** mcp 2.x renamed `FastMCP` to `MCPServer`; `scene_service/mcp_tools.py` and `robonix-api` use the v1 API and now fail at import without the pin.

**`sim/stop.sh` no longer kills processes it does not own.** It matched `robonix-atlas|robonix-executor|...` across the whole machine, so running it took down an unrelated Robonix deployment on the same host. It now scopes by working directory and cmdline ancestry under the Webots checkout.

## Replica (ConceptGraphs scorer, live replay through Scene)

| metric | Scene lite (was) | Scene + DualMap |
|---|---|---|
| objects n=1 (mIoU / mAcc / F-mIoU) | 4.6 / 18.3 / 5.4 | 12.8 / 29.5 / 18.4 |
| objects n=6 | 9.9 / 22.0 / 14.2 | 23.4 / 31.7 / 54.8 |
| relations, ReplicaSSG R@1 (obj / pred / rel) | 0.15 / 0.12 / 0.12 | 0.34 / 0.32 / 0.32 |

For reference on the same scorer: DualMap standalone with its general vocabulary is 13.1 / 27.6 / 19.9 at n=1, and the offline paper-recipe ConceptGraphs (SAM ViT-H + CLIP ViT-H-14) reaches 39.8 / 21.4 mAcc / F-mIoU across 4 scenes. Scene's live numbers track DualMap standalone, which is the point — the backend is reproduced, not reinvented.

## What this comparison is, and is not

The in-repo backend runs its `lite` profile in every row below — YOLO-World detection, MobileSAM segmentation, CLIP ViT-B-32 features — not the ConceptGraphs paper's configuration, which segments whole frames with SAM ViT-H and encodes with CLIP ViT-H-14, offline. A `full` profile with the larger weights exists in the repository, but a 6 GiB card cannot hold it alongside the simulator, and offline batch processing is not how a robot runs.

So these rows compare **our lightweight online loop** against **DualMap's default configuration**, on the hardware this deployment uses. That answers whether switching backends is worth it here. It does **not** establish which underlying method is stronger, and the effect of the smaller segmentation and feature models on the in-repo backend's numbers has not been measured — ruling it out would need a third arm running the `full` profile on a larger card.

## How the Webots rows are measured

The comparison is only worth reading if the ruler is stated, so:

**Ground truth comes out of the world file, not off a clipboard.** `export_webots_scene_truth.py` reads `office.wbt` and turns its nodes into truth objects through the 62 semantic types in `testing/fixtures/webots_scene_benchmark.json` (`Sofa → couch`, 2.0 x 0.9 x 0.8 m, association radius 0.75 m, and so on). Change the world and the truth follows.

**Only what the robot actually saw is scored.** An RGB-D sweep produces `visible_truth_ids`; a truth never in view is not counted against recall, or the score measures the tour rather than the perception. The office tour covers 1.0 (all 51 truths seen), so recall here is clean.

**Association is geometric, deliberately blind to the label.** A Hungarian one-to-one assignment over hard admissibility gates (a per-type ground-plane radius, 0.5–0.9 m; bbox volume ratio >= 0.03) maximizes match count, then prefers semantically correct labels, then minimizes normalized XY distance. Keeping the label out of association is the point: a detection in the right place with the wrong class must count as a label error, not as a miss.

**False positives are split, because they need different cures.**

| | |
| --- | --- |
| TP | matched one-to-one with a visible truth |
| FN | a visible truth with no match |
| **duplicate** | unmatched, but inside some truth's association radius — one object counted twice |
| **ghost** | unmatched and near no truth — something invented |

Merging fixes duplicates; filtering fixes ghosts. A single FP number hides which of the two a change moved, which is exactly what went wrong in the rounds below.

**Labels are judged for meaning** (see the judge script above), and the gates in the benchmark are precision >= 0.75, recall >= 0.7, F1 >= 0.72, label accuracy >= 0.8, duplicate rate <= 0.1, ghost rate <= 0.05. **None of them passes yet.**

**A round only counts when localization stayed healthy.** Scene can only register what the robot drove past. One round scored F1 0.342 off 25 objects; the cause was a localization the mapping service had marked `suspect`, four of five nav2 goals failing, and a robot that barely moved — a navigation failure being read as a perception result. Rounds now require 4 of 5 goals to succeed with at most one `suspect` reading during the tour, and are discarded otherwise. **The two backends are run alternately**, not one after the other, so any drift in the machine lands on both sides equally.

## Webots office: the backend swapped, nothing else

Every row: the office world, the same saved map, the same five nav2 goals, the same scorer, one image (`robonix-scene-dualmap`, which carries both backends' weights), and only `perception.backend` changed in the manifest. A round counts only if 4 of 5 goals succeeded with at most one `suspect` localization reading; invalid rounds were retried, not reported.

**Before two fixes landed the same evening** — the saved map's `free_thresh` served unexplored cells to nav2 as free floor (syswonder/service-map-rbnx#24), and tour goals were picked with a nine-point clearance check that let a 0.25 m gap through. Both backends ran on that map, so the rows are comparable with each other and not with the group below.

| backend | rounds | TP | FN | dup | ghost | precision | recall | **F1** | label acc |
|---|---|---|---|---|---|---|---|---|---|
| `concept_graphs` (Scene as on dev) | 2 | 28 / 22 | 23 / 29 | 4 / 2 | 3 / 2 | 0.800 / 0.846 | 0.549 / 0.431 | **0.651 / 0.571** | 0.536 / 0.591 |
| `dualmap`, DualMap's own 101-name home list | 2 | 30 / 31 | 21 / 20 | 16 / 20 | 21 / 24 | 0.448 / 0.413 | 0.588 / 0.608 | 0.508 / 0.492 | 0.367 / 0.452 |
| `dualmap` + the 40-name office list | 1 | 27 | 24 | 12 | 14 | 0.509 | 0.529 | 0.519 | **0.815** |

**After** (unknown cells stay unknown; goals need 0.9 m of real clearance):

| backend | rounds | TP | FN | dup | ghost | precision | recall | **F1** | label acc |
|---|---|---|---|---|---|---|---|---|---|
| `concept_graphs` | 2 | 36 / 26 | 15 / 25 | 5 / 5 | 5 / 7 | 0.783 / 0.684 | 0.706 / 0.510 | **0.742 / 0.584** | 0.639 / 0.731 |
| `dualmap` + office list | 1 | 35 | 16 | 5 | 8 | 0.729 | 0.686 | **0.707** | 0.600 |

What the two tables say, stated plainly:

- On the broken map the old Scene was far more precise (0.82 vs 0.43) and DualMap only won recall. Both were working inside a wrong world model; the old one was simply more conservative in it.
- On the fixed map they are level — `concept_graphs` 0.742 / 0.584 with a 0.16 swing between rounds, `dualmap` 0.707 on one round. **This PR does not claim DualMap scores higher on object F1.** Three rounds a side would be needed to say either way.
- The vocabulary is the largest single lever in the PR: label accuracy 0.41 → **0.82** for DualMap, above the old Scene's 0.56. DualMap is better at saying *what* a thing is; whether it is better at not inventing things is not settled.
- Fixing the map lifted both backends (0.61 → 0.66 and 0.52 → 0.71). Perception is scored against a world model, and a wrong one drags every backend down together.

## What each backend costs

Not measured until this evening, and the reason DualMap was chosen in the first place. Measured on the workpc (RTX 2060, 6 GB; 30.6 GB host RAM), sampled every 5 s on the same tour. "RSS" is the scene container's memory as `docker stats` reports it — host RAM, not GPU memory; GPU memory is its own column, and both backends fit the 6 GB card.

| backend | CPU (median, 1 core = 100%) | container RAM, start → 5 min | GPU memory used (card) | GPU util |
|---|---|---|---|---|
| `concept_graphs` | 94% (≈ one core) | **1.2 GB → 22.5 GB** | 3.9 GB | 37% |
| `dualmap` + office list | **24%** | **2.2 GB → 2.6 GB (flat)** | 3.6 GB | 18% |

The steady state differs modestly and traceably: DualMap voxel-downsamples an object's cloud after every fusion, while the in-repo backend does so only once an object passes a per-object cap, so the container settles at 2.3 GB against 2.8 GB. CPU differs more, and also traceably — the in-repo backend runs detection, segmentation and encoding every tick; DualMap runs them only after 0.15 m of travel or 5 degrees of turn.

Separately, and worth reporting because of what it would mean on a robot: on one tour the in-repo backend's resident set went from 2.8 GB to 22.0 GB within one sampling interval, stayed there two minutes, then dropped to 0.9 GB as the container restarted. DualMap showed nothing comparable on the same route. This is an excursion, not gradual growth, and it is not being called a leak: the cause is not localized yet (the periodic merge pass, which concatenates two clouds and re-processes them through an optional DBSCAN, is the leading suspicion). Repeated rounds are running to establish whether it reproduces.

The scene status page now shows these live, with the backend's name (`/cost`).

## Validation

- `system/scene/tests/test_perception_dualmap.py` — map-entry shaping, the manifest vocabulary, keyframe gating, uid preservation across merges, and the lifecycle plumbing.
- `system/scene/tests/test_capabilities.py` — backend routing and config validation (an unknown key under `perception.dualmap` fails at load).
- Live Webots runs above (each a fresh boot, static relocalization, five nav2 goals, one score); live Replica room0 replays for the Replica table.

## Notes

- The DualMap image (`docker/Dockerfile.dualmap`) clones the pinned commit and downloads FastSAM-s and MobileCLIP-S2 with size checks; Scene's own code is still bind-mounted.
- `examples/webots/robonix_manifest.scene-eval.dualmap.nav.yaml` is the manifest these runs used; the `.dualmap.yaml` variant is the same without nav2.
- The benchmark harness that drove earlier rounds (runner, sweep, report renderer, its tests and fixtures) is out of this PR; the scorer and the ground-truth export that produced every number here stay.

## Screenshot

Scene's live view on the DualMap backend in the Webots office world. Left is the 2D map with tracked objects, centre the 3D view with object boxes and relation edges, right the live RGB and depth the backend is reading.
scene-dualmap-3d-view


enkerewpo:
## Test report

Run on a live Webots office session (TIAGo, `robonix_manifest.scene-eval.dualmap.yaml`, `robonix-scene-dualmap:repo`, RTX 2060), driven on a manual tour.

### Automated

`pytest system/scene/tests` — **287 passed, 6 skipped, 0 failed**, after merging `dev`.

Four conflicts, all dependency manifests, all the same change arriving from both sides: `dev` pinned mcp / fastmcp / grpcio / protobuf while this branch was open. Resolved to `dev` throughout — same intent, expressed more precisely. `uv lock --check` is clean against the resolved pyprojects.

### End to end

DualMap loads and runs: 40-name vocabulary from the manifest, `device=cuda`, keyframe gate active. Over one tour the local map went 29 → 138 tracks and the registry settled at **64 objects**, named from the deployment vocabulary: couch, potted_plant, vase, trash_can, bench, table, lamp x3, cabinet, chair x4, monitor, keyboard, laptop, shelf, cushion, picture. All 13 scene capabilities register, including `delete_object`, `flush_objects`, `update_object_label`.

The drop counters behave as documented and are worth reading during a run: with the robot stationary every track was dropped `few-points`; rotating in place moved most drops to `unmapped` (the occupancy gate — an object on ground the map has not observed); only after the robot actually traversed the space did objects survive.

### Fixed in this PR

**`delete` and `flush` never reached the global map.** DualMap keeps two: the local map holds what the robot is looking at, and a track that becomes stable is promoted to the global map, keeping its uid. `_tick_locked` rebuilds the scene from the union of both. Both operator hooks edited only the local map, so with `global_map` on, deleting an object removed it from the local map and the registry and the next tick reinstated it from the global one — the operator deletes it and it comes back. Flush had the same hole and left every promoted track in place. `global_map` is opt-in and off by default, so this only bit a deployment that turned it on — which is also the one where nobody would think to look, since delete works everywhere else. Three tests pin it.

**A stale assertion on the confidence fallback.** `_track_confidence` stopped inventing a number when nothing has measured one; the GlobalObject shape test still asserted the old `0.5`. It is the one case that reaches the fallback.

## Two things a reviewer should know

### The benchmark in this PR cannot be run

`6b32852f` ("keep the PR to the backend and what scores it") removed the harness — `run_webots_scene_sweep.py`, `webots_scene_motion.py`, `webots_scene_visibility.py`, the report renderer — but left `testing/run_webots_scene_benchmark.sh`, which still `docker cp`s the first three. It fails on the first one. `dev` does not carry them either.

`evaluate_webots_scene.py` survived, but it alone is not enough: it takes the map-to-world alignment (`truth_alignment`: translation and yaw) from `--visibility-file`, which is what the deleted sweep produced. Without it the truth stays in world coordinates while Scene's objects are in map coordinates. Scored that way this run gives F1 0.048 with **zero** matched objects while the labels are plainly right — that number measures the missing alignment, not DualMap, and should not be read as a result.

So the numbers in the PR body are not reproducible from the PR as it stands. Either restore the three helpers, or say in the description where they live.

Separately, the default config gate (`candidate`) cannot pass on this branch or on `dev`: it requires `clip_rerank_*` and `surface_snap` to have been applied, and neither branch implements them — `grep` finds no reference in `scene_service`. `CONFIG_GATE_MODE=baseline` is what runs today.

### Duplicate objects are still the open problem

On the in-repo backend one continuous run of office shelving comes back as six objects — `cabinet`, `shelf`, `cabinet`, `shelf` plus `shelf_009..012` within two metres of each other. The merge criterion is point-cloud overlap, and two adjacent segments of one long object do not overlap each other, so no distance threshold reaches them; `max_merge_dist_m` never gets a say.

The `association` block the eval manifests carry (`same_class_centroid_max_m`, `exact_duplicate_*`, `coobserved_duplicate_*`, ~20 knobs written for exactly this) is **not read by any code on this branch, on #248, or on `dev`** — Scene logs those keys as "ignored (not implemented)" at startup. Its implementation is on `scene-object-management-177` / `scene-benchmark-harness`, neither merged.

DualMap's global map merges by top-down 2D box overlap across classes, which is the mechanism aimed at this, and it is the path where the delete/flush bug above lived. It stays off by default, per the PR.

system/scene/config.spec
```diff
@@ -0,0 +1,192 @@
...
+  # optional integer port, default: 50107 (environment fallback
+  # SCENE_WEB_PORT). 0 disables the web interface.
+  web_port: 50107
+
+  # ── perception ───────────────────────────────────────────────────────────
+
+  perception:
+
+    # string, default: lite (environment fallback SCENE_PROFILE).
+    # lite      the 2060-class model set, about 4 GB
```
enkerewpo: never use hardcode model name like 2060

system/scene/scene_service/ingest/cg_export.py
enkerewpo, line 1: remove this helper for simplicity

system/scene/scene_service/ingest/geometry_gates.py
```diff
@@ -0,0 +1,89 @@
+# SPDX-License-Identifier: MulanPSL-2.0
+"""Things the world cannot be, decided by the map rather than by a threshold.
```
enkerewpo, line 2: should rewrite these descriptions

enkerewpo:
OK to merge with minor changes

Issue #229, opened 2026-08-11
Scene Web Annotation API allows unauthenticated CRUD operations


0XF1-sec:
Robonix Scene Web exposes Annotation API endpoints without application-layer authentication or authorization checks.

An unauthenticated client can directly read, create, update, and delete map annotations through the following endpoints:

GET    /api/annotations
POST   /api/annotations
PUT    /api/annotations/{annotation_id}
DELETE /api/annotations/{annotation_id}

Even when a clearly invalid Bearer Token is supplied, the server still returns 200 OK and processes the request.

This allows any network client that can reach the Scene Web service to tamper with annotation data such as rooms, points of interest, coordinates, and headings.

Affected Component
Component: system/scene
system/scene/scene_service/web.py
Function: make_app()
Affected handlers:
annotations_list
annotations_create
annotations_update
annotations_delete
Source Location

Repository:

https://github.com/syswonder/robonix

Affected source file:

https://github.com/syswonder/robonix/blob/dev/system/scene/scene_service/web.py

Relevant code area:

async def annotations_list(_request):
    if anno_store is None:
        return _anno_error(503, "annotation store unavailable")
    return JSONResponse({"ok": True, "annotations": anno_store.list_json()})

async def annotations_create(request):
    ...
    ann = anno_store.create(kind=kind, name=name, points=points, theta=theta)
    return JSONResponse({"ok": True, "annotation": ann.to_json()})

async def annotations_update(request):
    ...
    ann = anno_store.update(...)
    return JSONResponse({"ok": True, "annotation": ann.to_json()})

async def annotations_delete(request):
    ...
    if not anno_store.delete(ann_id):
        return _anno_error(404, f"unknown annotation id {ann_id!r}")
    return JSONResponse({"ok": True})

The route registration directly exposes the Annotation CRUD endpoints:

Route("/api/annotations", annotations_list, methods=["GET"]),
Route("/api/annotations", annotations_create, methods=["POST"]),
Route("/api/annotations/{annotation_id}", annotations_update, methods=["PUT"]),
Route("/api/annotations/{annotation_id}", annotations_delete, methods=["DELETE"]),

The implementation does not check:

Authorization header;
Bearer Token validity;
Cookie or session state;
user identity;
role permission;
annotation ownership;
resource-level access control.
Environment

The issue was reproduced in a minimal Scene Web deployment.

[...]

PR #245, opened 2026-09-09
fix(scene): survive malformed upstream messages and keep the web UI local

### LINKED ISSUE #226: [BUG] Validate upstream message structure consistency to prevent malformed OccupancyGrid/Image from causing interface-level denial of service
# fix(scene/web): Validate upstream message structure consistency to prevent malformed OccupancyGrid/Image from causing interface-level denial of service

## Overview

`scene web` reads upstream ROS2 messages and renders them on the web page by directly calling `np.frombuffer(...).reshape(h, w)` on `msg.data` without validating that the `data` length matches `width * height`. An attacker only needs to publish a structurally inconsistent but type-valid message to ROS2 topics such as `/map`, causing `GET /api/state` and `GET /api/camera` to consistently return `500 Internal Server Error`, resulting in an interface-level denial of service.

---

### 1.1 OccupancyGrid sink

**File**: [web.py](file:///e:/robonix-dev/system/scene/scene_service/web.py#L254-L254)

```python
arr = np.frombuffer(bytes(msg.data), dtype=np.int8).reshape(h, w)
```

- **Triggering route**: `GET /api/state` → `_state_payload()` → `_occupancy_payload(hub)`
- **Data source**: `hub.latest("occupancy_grid")` → `_LatestSlot._msg` (stores original `nav_msgs/msg/OccupancyGrid`)
- **Crash condition**: `len(msg.data) != info.width * info.height`
- **Exception type**: `ValueError: cannot reshape array of size N into shape (h, w)`

Call chain:

```
GET /api/state (route, L479)
  → _state_payload(registry, hub, sg_store) (L373)
    → _occupancy_payload(hub) (L429)
      → hub.latest("occupancy_grid") (L236)        ← retrieve malformed message
      → np.frombuffer(...).reshape(h, w) (L254)    ← SINK: raises ValueError
```

### 1.2 RGB Image sinks

**File**: [web.py](file:///e:/robonix-dev/system/scene/scene_service/web.py#L297-L306)

| Line | encoding | target shape for reshape | required data length |
|------|----------|---------------------------|----------------------|
| L297 | `rgb8` | `(h, w, 3)` | `h * w * 3` |
| L299 | `bgr8` | `(h, w, 3)` | `h * w * 3` |
| L301 | `rgba8` | `(h, w, 4)` | `h * w * 4` |
| L304 | `bgra8` | `(h, w, 4)` | `h * w * 4` |
| L306 | `mono8` | `(h, w)` | `h * w` |

- **Triggering route**: `GET /api/camera` → `_camera_payload(hub)` → `_image_to_png_b64(msg, kind="rgb")`
- **Data source**: `hub.latest("rgb")` → `_LatestSlot._msg` (stores original `sensor_msgs/msg/Image`)
- **Crash condition**: `len(msg.data) != h * w * channels`

### 1.3 Depth Image sinks

**File**: [web.py](file:///e:/robonix-dev/system/scene/scene_service/web.py#L312-L315)

| Line | encoding | dtype | target shape for reshape | required data length |
|------|----------|-------|---------------------------|----------------------|
| L312 | `32fc1` / `32FC1` | `np.float32` | `(h, w)` | `h * w * 4` bytes |
| L315 | `16uc1` / `16UC1` | `np.uint16` | `(h, w)` | `h * w * 2` bytes |

- **Triggering route**: `GET /api/camera` → `_camera_payload(hub)` → `_image_to_png_b64(msg, kind="depth")`
- **Data source**: `hub.latest("depth")` → `_LatestSlot._msg`

---

### 2.1 Path for messages entering the hub

```
ROS2 Publisher (attacker or faulty node)
  → /map topic
  → SubscribersHub._subscribe() creates subscription
  → rclpy callback _cb(msg)  [ros_subscribers.py:289]
  → _LatestSlot.write(msg)    [ros_subscribers.py:97]
  → _LatestSlot._msg = msg    ← malformed message is stored as‑is, no validation
```

Key code — [ros_subscribers.py](file:///e:/robonix-dev/system/scene/scene_service/ingest/ros_subscribers.py#L289-L290):

```python
def _cb(msg: Any, _slot: _LatestSlot = slot, _k: str = _kind) -> None:
    _slot.write(msg)   # ← no validation, stores directly
```

Key code — [ros_subscribers.py](file:///e:/robonix-dev/system/scene/scene_service/ingest/ros_subscribers.py#L97-L101):

```python
def write(self, msg: Any) -> None:
    with self._lock:
        self._msg = msg          # ← stored without checking structure
        self._stamp_unix = time.time()
        self._count += 1
```

### 2.2 Web route reading from hub

```
HTTP GET /api/state
  → Starlette route handler (web.py:479)
  → _state_payload(registry, hub, sg_store) (web.py:373)
  → _occupancy_payload(hub) (web.py:229)
  → hub.latest("occupancy_grid") (web.py:236)  ← retrieve malformed msg
  → np.frombuffer(bytes(msg.data), ...).reshape(h, w)  ← SINK: ValueError
  → Exception uncaught, Starlette returns 500
```

### 2.3 State persistence

Because `_LatestSlot` keeps only the latest message, and `/map` uses `TRANSIENT_LOCAL` QoS:

1. Even if the malformed publisher stops, `scene` internally retains the last malformed message.
2. The latching semantics of `TRANSIENT_LOCAL` mean that new subscribers may also receive this malformed message.
3. The interface only recovers when a new valid message overwrites `slot._msg`.

This gives the vulnerability a **persistent denial-of-service** characteristic.

---

### 3.1 Deployment‑level PoC (validated on a real Robonix deployment)

**Environment**:
- Simulation container: `robonix_tiago_sim`
- `scene web` listening on: `http://127.0.0.1:50107/`
- Map topic: `/map` (`nav_msgs/msg/OccupancyGrid`)
- RGB topic: `/head_front_camera/rgb/image_raw` (`sensor_msgs/msg/Image`)

**Step 1 — Baseline verification**:

```bash
# All interfaces work normally in clean state
curl --noproxy '*' -i http://127.0.0.1:50107/
curl --noproxy '*' -i http://127.0.0.1:50107/api/state
curl --noproxy '*' -i http://127.0.0.1:50107/api/camera

# Expected: all 200 OK
```

**Step 2 — Poison /map (triggers /api/state failure)**:

```bash
# Terminal A: publish malformed OccupancyGrid to real /map topic
# width=4, height=4 but data has only 3 elements
docker exec robonix_tiago_sim bash -lc \
  'source /opt/ros/humble/setup.bash && \
   ros2 topic pub -r 5 /map nav_msgs/msg/OccupancyGrid \
   "{header: {frame_id: map}, info: {resolution: 0.05, width: 4, height: 4, origin: {position: {x: 0.0, y: 0.0, z: 0.0}, orientation: {w: 1.0}}}, data: [0, 0, 0]}"'

# Terminal B: send ordinary HTTP request within contamination window
curl --noproxy '*' -i http://127.0.0.1:50107/api/state
```

**Expected result**:

```
HTTP/1.1 500 Internal Server Error
```

`scene` logs also show:

```text
File "/scene/scene_service/web.py", line 254, in _occupancy_payload
  arr = np.frombuffer(bytes(msg.data), dtype=np.int8).reshape(h, w)
ValueError: cannot reshape array of size 3 into shape (4,4)
```

**Step 3 — Poison RGB topic (triggers /api/camera failure)**:

```bash
# Terminal A: publish malformed Image to real RGB topic
# width=10, height=10, encoding=rgb8 but data only 20 bytes (should be 300)
docker exec robonix_tiago_sim bash -lc "
  source /opt/ros/humble/setup.bash
  export RMW_IMPLEMENTATION=rmw_zenoh_cpp
  timeout 8s ros2 topic pub -r 10 /head_front_camera/rgb/image_raw sensor_msgs/msg/Image \
  '{
    header: {frame_id: head_front_camera_rgb_optical_frame},
    height: 10,
    width: 10,
    encoding: rgb8,
    is_bigendian: 0,
    step: 30,
    data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
  }'
"

# Terminal B: send ordinary HTTP request
curl --noproxy '*' -i http://127.0.0.1:50107/api/camera
```

**Expected result**:

```
HTTP/1.1 500 Internal Server Error
```

Logs:

```text
ValueError: cannot reshape array of size 20 into shape (10,10,3)
```

### 3.2 Component‑level PoC

```python
"""Component‑level PoC: demonstrate the inherent flaw in web.py.
Load the real web.py, replace external dependencies with stubs,
inject malformed messages into the hub in‑process, then send ordinary HTTP requests."""
import asyncio
import httpx
import uvicorn
from types import SimpleNamespace

# --- stub external dependencies ---
import sys

# stub robonix_api and scene_service.state
class _StubRegistry:
    _objects = {}
    _surfaces = {}

class _StubHub:
    """Mock SubscribersHub that can inject malformed messages."""
    def __init__(self):
        self._slots = {}
    def inject(self, kind, msg):
        self._slots[kind] = (msg, 1700000000.0, 1)
    def has(self, kind):
        return kind in self._slots
    def latest(self, kind):
        return self._slots.get(kind, (None, 0.0, 0))

# stub modules
stub_state = SimpleNamespace(ObjectRegistry=_StubRegistry)
sys.modules['scene_service'] = SimpleNamespace()
sys.modules['scene_service.state'] = stub_state

# --- load real web.py ---
from scene_service.web import make_app, _StubRegistry  # may differ by layout

# Construct malformed OccupancyGrid
malformed_grid = SimpleNamespace(
    info=SimpleNamespace(
        width=4, height=4,
        resolution=0.05,
        origin=SimpleNamespace(
            position=SimpleNamespace(x=0.0, y=0.0, z=0.0),
        ),
    ),
    data=[0, 0, 0],  # only 3 elements, should be 4*4=16
)

# Construct malformed RGB Image
malformed_rgb = SimpleNamespace(
    height=10, width=10,
    encoding="rgb8",
    data=bytes(20),  # only 20 bytes, should be 10*10*3=300
    header=SimpleNamespace(
        stamp=SimpleNamespace(sec=0, nanosec=0)
    ),
)

hub = _StubHub()
registry = _StubRegistry()

# Inject malformed messages
hub.inject("occupancy_grid", malformed_grid)
hub.inject("rgb", malformed_rgb)

app = make_app(registry=registry, hub=hub)

config = uvicorn.Config(app, host="127.0.0.1", port=18099, log_level="warning")
server = uvicorn.Server(config)

async def run_test():
    """Start server and send HTTP requests."""
    import threading
    t = threading.Thread(target=server.run, daemon=True)
    t.start()
    await asyncio.sleep(1.0)

    async with httpx.AsyncClient() as client:
        # / should be normal
        r = await client.get("http://127.0.0.1:18099/")
        print(f"GET / status={r.status_code}")

        # /api/state should be 500 (malformed occupancy grid)
        r = await client.get("http://127.0.0.1:18099/api/state")
        print(f"GET /api/state status={r.status_code}")

        # /api/camera should be 500 (malformed rgb image)
        r = await client.get("http://127.0.0.1:18099/api/camera")
        print(f"GET /api/camera status={r.status_code}")

asyncio.run(run_test())

# Expected output:
# GET / status=200
# GET /api/state status=500
# GET /api/camera status=500
```

---

### 4.1 Design Principles

1. **Validation at entry**: Validate `data` length consistency with metadata before `reshape`.
2. **Graceful degradation**: On validation failure, return `None` (skip rendering) rather than raising an exception up to the route layer.
3. **Route‑level fallback**: Add `try/except` at the route layer to ensure any unexpected exception returns a structured error instead of a raw 500.

#### 4.2.1 Add validation helper functions

Insert before `_shorten_id` at [web.py](file:///e:/robonix-dev/system/scene/scene_service/web.py#L216-L218):

```python
def _check_grid_consistency(msg: Any, w: int, h: int) -> bool:
    """Verify OccupancyGrid data length matches width*height."""
    expected = w * h
    actual = len(msg.data)
    if actual != expected:
        log.warning(
            "[web] occupancy: data length %d != width*height %d*%d=%d; "
            "skipping render",
            actual, w, h, expected,
        )
        return False
    return True


def _check_image_consistency(msg: Any, w: int, h: int, channels: int) -> bool:
    """Verify Image data byte length matches width*height*channels."""
    expected = w * h * channels
    actual = len(msg.data)
    if actual != expected:
        log.warning(
            "[web] image(%s): data length %d != %dx%dx%d=%d; "
            "skipping render",
            msg.encoding, actual, w, h, channels, expected,
        )
        return False
    return True
```

#### 4.2.2 Fix `_occupancy_payload`

At [web.py:248-254](file:///e:/robonix-dev/system/scene/scene_service/web.py#L248-L254), replace:

```python
    info = msg.info
    w, h = int(info.width), int(info.height)
    if w == 0 or h == 0:
        return None
    # nav_msgs/OccupancyGrid data is row-major bottom-up int8 in
    # [-1, 100]: -1 unknown, 0 free, 100 occupied. Render as grayscale:
    # unknown=128 (mid), free=240 (almost white), occupied=20 (almost black).
    arr = np.frombuffer(bytes(msg.data), dtype=np.int8).reshape(h, w)
```

with:

```python
    info = msg.info
    w, h = int(info.width), int(info.height)
    if w == 0 or h == 0:
        return None
    if not _check_grid_consistency(msg, w, h):
        return None
    # nav_msgs/OccupancyGrid data is row-major bottom-up int8 in
    # [-1, 100]: -1 unknown, 0 free, 100 occupied. Render as grayscale:
    # unknown=128 (mid), free=240 (almost white), occupied=20 (almost black).
    arr = np.frombuffer(bytes(msg.data), dtype=np.int8).reshape(h, w)
```

#### 4.2.3 Fix `_image_to_png_b64`

At [web.py:289-317](file:///e:/robonix-dev/system/scene/scene_service/web.py#L289-L317), for each encoding branch add a channel count and validation.

Replace:

```python
    h, w = int(msg.height), int(msg.width)
    if h == 0 or w == 0:
        return None
    enc = (msg.encoding or "").lower()
    arr: Any = None
    out_mode = "RGB"
    if kind == "rgb":
        if enc == "rgb8":
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w, 3)
        elif enc == "bgr8":
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w, 3)[:, :, ::-1]
        elif enc == "rgba8":
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w, 4)[:, :, :3]
        elif enc == "bgra8":
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w, 4)[:, :, :3][:, :, ::-1]
        elif enc == "mono8":
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w)
            out_mode = "L"
        else:
            return None
    else:  # depth
        if enc in ("32fc1", "32FC1"):
            raw = np.frombuffer(bytes(msg.data), dtype=np.float32).reshape(h, w)
        elif enc in ("16uc1", "16UC1"):
            raw = np.frombuffer(bytes(msg.data), dtype=np.uint16).reshape(h, w).astype(np.float32) / 1000.0
        else:
            return None
```

with:

```python
    h, w = int(msg.height), int(msg.width)
    if h == 0 or w == 0:
        return None
    enc = (msg.encoding or "").lower()
    arr: Any = None
    out_mode = "RGB"
    if kind == "rgb":
        if enc == "rgb8":
            if not _check_image_consistency(msg, w, h, 3): return None
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w, 3)
        elif enc == "bgr8":
            if not _check_image_consistency(msg, w, h, 3): return None
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w, 3)[:, :, ::-1]
        elif enc == "rgba8":
            if not _check_image_consistency(msg, w, h, 4): return None
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w, 4)[:, :, :3]
        elif enc == "bgra8":
            if not _check_image_consistency(msg, w, h, 4): return None
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w, 4)[:, :, :3][:, :, ::-1]
        elif enc == "mono8":
            if not _check_image_consistency(msg, w, h, 1): return None
            arr = np.frombuffer(bytes(msg.data), dtype=np.uint8).reshape(h, w)
            out_mode = "L"
        else:
            return None
    else:  # depth
        if enc in ("32fc1", "32fc1"):
            if not _check_image_consistency(msg, w, h, 4): return None
            raw = np.frombuffer(bytes(msg.data), dtype=np.float32).reshape(h, w)
        elif enc in ("16uc1", "16UC1"):
            if not _check_image_consistency(msg, w, h, 2): return None
            raw = np.frombuffer(bytes(msg.data), dtype=np.uint16).reshape(h, w).astype(np.float32) / 1000.0
        else:
            return None
```

#### 4.2.4 Add exception fallback at route layer

At [web.py:479-480](file:///e:/robonix-dev/system/scene/scene_service/web.py#L479-L480) and [web.py:493-494](file:///e:/robonix-dev/system/scene/scene_service/web.py#L493-L494), replace:

```python
    async def state(_request) -> JSONResponse:
        return JSONResponse(_state_payload(registry, hub, sg_store))

    ...

    async def camera_state(_request) -> JSONResponse:
        return JSONResponse(_camera_payload(hub))
```

with:

```python
    async def state(_request) -> JSONResponse:
        try:
            return JSONResponse(_state_payload(registry, hub, sg_store))
        except Exception:
            log.exception("[web] /api/state failed")
            return JSONResponse({"error": "internal_error"}, status_code=500)

    ...

    async def camera_state(_request) -> JSONResponse:
        try:
            return JSONResponse(_camera_payload(hub))
        except Exception:
            log.exception("[web] /api/camera failed")
            return JSONResponse({"error": "internal_error"}, status_code=500)
```

### 4.3 Summary of Changes

| File | Location | Change |
|------|----------|--------|
| `web.py` | before L216 | Add `_check_grid_consistency` + `_check_image_consistency` helper functions |
| `web.py` | L248-254 | `_occupancy_payload`: call `_check_grid_consistency` before reshape |
| `web.py` | L289-317 | `_image_to_png_b64`: add channel‑based validation for each encoding branch |
| `web.py` | L479-480 | `state` route: add try/except fallback |
| `web.py` | L493-494 | `camera_state` route: add try/except fallback |

### 4.4 Behaviour After Fix

| Scenario | Before Fix | After Fix |
|----------|------------|-----------|
| Malformed OccupancyGrid in hub | `GET /api/state` → 500 | `GET /api/state` → 200, `occupancy: null` (warning logged) |
| Malformed RGB Image in hub | `GET /api/camera` → 500 | `GET /api/camera` → 200, `rgb: null` (warning logged) |
| Valid message in hub | `GET /api/state` → 200 | `GET /api/state` → 200 (no change) |
| Other unexpected exception | 500 + raw stack trace | 500 + `{"error": "internal_error"}` + full stack trace logged |

---

## 5. Impact Assessment

- **Vulnerability type**: Input consistency validation missing / interface‑level denial of service
- **Attack surface**: Any entity that can publish messages to ROS2 topics subscribed by `scene`
- **Ease of exploitation**: Low — a single `ros2 topic pub` command is sufficient
- **Impact scope**: `/api/state` and `/api/camera` endpoints persistently return 500
- **Recovery cost**: Requires a new valid message to overwrite the slot, or restart of the `scene` service
- **CVSS assessment**: AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:L → Medium (3.7)

---

## 6. Verification Commands

```bash
# 1. Verify that after the fix the endpoints work normally in a clean state
curl --noproxy '*' -i http://127.0.0.1:50107/api/state
# Expected: 200

# 2. Verify that malformed messages no longer cause a 500
docker exec robonix_tiago_sim bash -lc \
  'source /opt/ros/humble/setup.bash && \
   ros2 topic pub --once /map nav_msgs/msg/OccupancyGrid \
   "{header: {frame_id: map}, info: {resolution: 0.05, width: 4, height: 4, origin: {position: {x: 0.0, y: 0.0, z: 0.0}, orientation: {w: 1.0}}}, data: [0, 0, 0]}"'

curl --noproxy '*' -i http://127.0.0.1:50107/api/state
# Expected: 200, occupancy field should be null

# 3. Verify that a warning appears in the logs
rbnx logs -t scene | grep "data length"
# Expected: [web] occupancy: data length 3 != width*height 4*4=16; skipping render
```

### LINKED ISSUE #227: [BUG] Validate occupancy_grid metadata consistency to prevent resolution=0 / length mismatch from causing goal_near division‑by‑zero crashes
# fix(scene/mcp_tools): Validate occupancy_grid metadata consistency to prevent resolution=0 / length mismatch from causing goal_near division‑by‑zero crashes

## Overview

The `scene` MCP tool `goal_near`, when reading the `occupancy_grid` from the hub and performing BFS reachability search, directly trusts upstream message fields such as `resolution`, `width`, `height` and `data` without any consistency or boundary checks. An attacker needs only to publish a structurally self‑consistent but `resolution=0` `OccupancyGrid` to the ROS2 `/map` topic; any subsequent normal `goal_near(object_id=...)` capability call will then raise `ZeroDivisionError: float division by zero`, causing a capability‑level denial of service.

In addition, the same function contains an independent crash path: when `len(data) != width*height` (i.e. the same upstream contamination as in Issue #04), the BFS‑internal `np.reshape(h, w)` raises a `ValueError`. This is the same root cause as Issue #04 but triggered through a different entry point (MCP `goal_near` capability rather than the web API).

---

### 1.1 Primary vulnerability: resolution=0 division‑by‑zero (this is Issue #05)

**File**: [mcp_tools.py](file:///e:/robonix-dev/system/scene/scene_service/mcp_tools.py)

All locations where `resolution` (variable `res`) is used as a divisor – **3 places**:

| Line | Code | Dividend | Description |
|------|------|----------|-------------|
| L130 | `(_GOAL_NEAR_ROBOT_RADIUS_M + _GOAL_NEAR_CLEARANCE_M) / res` | `(0.3 + 0.4) = 0.7` | Inflation radius calculation; `res=0` directly raises `ZeroDivisionError` |
| L147 | `(wx - ogx) / res` | world x difference | World → grid coordinate conversion; `res=0` causes division by zero |
| L148 | `(wy - ogy) / res` | world y difference | Same for y direction |

All these divisions occur inside `_occupancy_bfs()`, which is called by `goal_near` at [L211-216](file:///e:/robonix-dev/system/scene/scene_service/mcp_tools.py#L211-L216):

```python
found = _occupancy_bfs(
    msg,
    float(target.pose.x),
    float(target.pose.y),
    approach_ang,
)
```

Simplified call stack:

```
MCP: robonix/system/scene/goal_near
  → goal_near(req)           [mcp_tools.py:L156]
    → _HUB.latest("occupancy_grid")          [L199]    ← retrieves poisoned msg
    → _occupancy_bfs(msg, tx, ty, ang)       [L211]
      → res = float(info.resolution)         [L120]    ← res=0.0
      → (0.3 + 0.4) / res                    [L130]    ← SINK: ZeroDivisionError
      or
      → (wx - ogx) / res                     [L147]    ← SINK: ZeroDivisionError
      → (wy - ogy) / res                     [L148]    ← SINK: ZeroDivisionError
```

### 1.2 Secondary vulnerability: data length mismatch causing reshape crash (mirror of Issue #04 in the MCP entry point)

**File**: [mcp_tools.py](file:///e:/robonix-dev/system/scene/scene_service/mcp_tools.py#L123-L123)

```python
grid = np.frombuffer(bytes(grid_msg.data), dtype=np.int8).reshape(h, w)
```

- **Trigger condition**: `len(msg.data) != width * height`
- **Exception type**: `ValueError: cannot reshape array of size N into shape (h, w)`
- **Difference from Issue #04**: #04 is triggered via `GET /api/state → _occupancy_payload()`; this one is triggered via `MCP goal_near → _occupancy_bfs()`. They are two independent crash paths caused by the same upstream contamination.

### 1.3 What is missing in `goal_near` pre‑checks?

Look at [L205](file:///e:/robonix-dev/system/scene/scene_service/mcp_tools.py#L205-L205):

```python
if msg is None or count == 0 or not msg.info.width or not msg.info.height:
```

This check only covers:
- `msg is None`
- `count == 0`
- `width == 0` (implicit truthiness)
- `height == 0` (implicit truthiness)

**It does NOT check**:
- `resolution > 0` (doesn't check zero, negative, NaN, Inf)
- `len(msg.data) == width * height`
- Numerical sanity of `resolution` (e.g. extremely small 1e‑300 could cause overflow)

---

### 2.1 Path for messages entering the hub (shared with Issue #04)

```
ROS2 Publisher (attacker / faulty node)
  → /map topic  [nav_msgs/msg/OccupancyGrid]
  → SubscribersHub._subscribe() callback _cb(msg)
  → _LatestSlot.write(msg)       [ros_subscribers.py:L97]
  → _LatestSlot._msg = msg       ← stored as‑is, without validating resolution or data length
```

**Key point**: There is no validation at the storage layer. Even a message with `resolution=0.0`, `data=[0]*3` (when width=4,height=4) will be accepted unconditionally as "latest map".

### 2.2 Trigger path for MCP capability invocation

```
Pilot / external tool caller
  → atlas discovers "robonix/system/scene/goal_near"
  → MCP protocol transfer (normal JSON‑RPC, all parameters perfectly valid)
  → goal_near(req: GoalNear_Request)    [mcp_tools.py:L156]
  → req.object_id = "scene.object.xxx"  ← completely normal, valid parameters
  → _HUB.latest("occupancy_grid")       ← retrieves poisoned msg from upstream
  → _occupancy_bfs(msg, ...)            ← enters BFS
  → L130 / L147 / L148                  ← SINK: ZeroDivisionError
  → exception not caught by goal_near
  → MCP: "mcp tool error: Error executing tool goal_near: float division by zero"
```

**Critical security properties**:
- The attacker **does not need** to craft a malformed MCP request body – JSON‑Schema validation passes completely.
- They only need to poison the upstream ROS2 `/map` topic first (no authentication; in typical deployments `/map` can be published by any node in the same network).
- Any subsequent legitimate `goal_near` call will crash.
- The crash persists until a valid `OccupancyGrid` overwrites the dirty message in the hub.

---

### 3.1 Deployment‑level PoC (validated on a real Robonix Webots deployment)

**Environment**:
- Research container: `robonix-audit-dev-v2`
- Simulation container: `robonix_tiago_sim`
- `scene web`: `http://127.0.0.1:50107/`
- Deployment method: `rbnx boot -f robonix_manifest.debug.cpu.scene.yaml`
- Key ROS2 topics:
  - `/map` → `nav_msgs/msg/OccupancyGrid`
  - `/head_front_camera/rgb/image_raw` → `sensor_msgs/msg/Image`
- `RMW_IMPLEMENTATION=rmw_zenoh_cpp`

#### Step 1 — Baseline verification (clean state)

Inside the research container:

```bash
source /opt/ros/humble/setup.bash
export PATH=/root/.cargo/bin:/root/.local/bin:$PATH
export ROBONIX_HOME=/tmp/robonix-home-audit
export RMW_IMPLEMENTATION=rmw_zenoh_cpp
cd /home/ysq/robonix/examples/webots

# 1. List scene tools to confirm registration
rbnx tools | grep scene
# Expected: at least scene.scene_list_objects, scene.scene_goal_near

# 2. Baseline list_objects to obtain a real object_id
rbnx ask "Call the scene tool list_objects once and print the returned object ids only."
# Objects returned in this test included:
#   - scene.object.potted_plant_001
#   - scene.object.picture_frame_001
#   - scene.object.robot_001

# 3. Baseline goal_near call (using a real object_id)
rbnx ask "Call the scene tool goal_near with object_id=scene.object.potted_plant_001 and print the raw tool result only."
# Expected: reachable=true or reachable=false + structured reason
# No ZeroDivisionError should appear
```

#### Step 2 — Poison the upstream `/map` (inject resolution=0.0 with self‑consistent data length)

In the **simulation container**, continuously publish the malformed map. Note: `data` must contain exactly `width*height` elements to avoid triggering the #04 reshape path first (we want to trigger the #05 division‑by‑zero path, not the #04 length path).

```bash
docker exec robonix_tiago_sim bash -lc \
  'source /opt/ros/humble/setup.bash && \
   ros2 topic pub -r 5 /map nav_msgs/msg/OccupancyGrid \
   "{header: {frame_id: map}, info: {resolution: 0.0, width: 4, height: 4, origin: {position: {x: 0.0, y: 0.0, z: 0.0}, orientation: {w: 1.0}}}, data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}"'
```

Key parameters:
```
resolution = 0.0   ← fatal field, triggers #05
width = 4
height = 4
data   = [0]*16    ← correct length 4*4=16, avoids the #04 path
```

#### Step 3 — Send a normal MCP call to trigger the crash

In the research container, use **exactly the same** legitimate parameters as in the baseline:

```bash
source /opt/ros/humble/setup.bash
export PATH=/root/.cargo/bin:/root/.local/bin:$PATH
export ROBONIX_HOME=/tmp/robonix-home-audit
export RMW_IMPLEMENTATION=rmw_zenoh_cpp
cd /home/ysq/robonix/examples/webots

rbnx ask "Call the scene tool goal_near with object_id=scene.object.potted_plant_001 and print the raw tool result only."
```

**Actual triggered result**:

```text
mcp tool error: Error executing tool goal_near: float division by zero
```

Full structured failure output:

```text
{"error":"mcp tool error: Error executing tool goal_near: float division by zero","output":"","success":false}
```

#### Step 4 — Check scene logs to confirm crash location

```bash
source /opt/ros/humble/setup.bash
export PATH=/root/.cargo/bin:/root/.local/bin:$PATH
export ROBONIX_HOME=/tmp/robonix-home-audit
export RMW_IMPLEMENTATION=rmw_zenoh_cpp
cd /home/ysq/robonix/examples/webots
rbnx logs -t scene -f 2>&1 | grep -A 10 "ZeroDivisionError\|goal_near"
```

Expected log stack trace (derived from line numbers):

```text
File "/scene/scene_service/mcp_tools.py", line 211, in goal_near
  found = _occupancy_bfs(msg, ...)
File "/scene/scene_service/mcp_tools.py", line 130, in _occupancy_bfs
  (_GOAL_NEAR_ROBOT_RADIUS_M + _GOAL_NEAR_CLEARANCE_M) / res
ZeroDivisionError: float division by zero
```

(If `max(1, ceil(...))` at L130 blows up first because `resolution=0`, the stack points to L130; if by chance the dividend is zero or the floating point does not raise and the code enters the loop, the stack will be at L147/L148.)

#### Step 5 — Recovery after evidence collection

First `Ctrl-C` to stop the `ros2 topic pub -r 5` in terminal A, then publish a normal map to overwrite the dirty slot:

```bash
docker exec robonix_tiago_sim bash -lc \
  'source /opt/ros/humble/setup.bash && \
   ros2 topic pub --once /map nav_msgs/msg/OccupancyGrid \
   "{header: {frame_id: map}, info: {resolution: 0.05, width: 4, height: 4, origin: {position: {x: 0.0, y: 0.0, z: 0.0}, orientation: {w: 1.0}}}, data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}"'
```

Then re‑execute the baseline `goal_near` to verify recovery.

---

### 3.2 Component‑level PoC (pure Python stub, no external dependencies)

Loads the real `mcp_tools.py` logic, replaces external dependencies (`ObjectRegistry`, `SubscribersHub`, `semantic_map_mcp`) with stubs. No need to install ROS2 or an MCP server.

```python
"""Component‑level PoC: verifies Issue #05 (resolution=0 causes goal_near division‑by‑zero crash)
and also the MCP‑entry mirror of Issue #04 (data length mismatch reshape crash).
No numpy required: we simulate the length‑check semantics of frombuffer().reshape() in pure Python.
"""
import sys
import math

# ────────────────────────────────────────────────────────────────
# 1. Mock scene_service.state.ObjectRegistry / SceneObject
# ────────────────────────────────────────────────────────────────
class FakePose:
    def __init__(self, x, y, z, yaw=0.0):
        self.x = x; self.y = y; self.z = z; self.yaw = yaw

class FakeSceneObject:
    def __init__(self, object_id, cls, x, y, z, yaw=0.0, is_robot=False, missing=False):
        self.object_id = object_id
        self.cls = cls
        self.pose = FakePose(x, y, z, yaw)
        self.is_robot = is_robot
        self.missing = missing
        self.last_seen = 1700000000.0

class FakeRegistry:
    def __init__(self, objects):
        self._objects = {o.object_id: o for o in objects}
    async def snapshot(self):
        return (dict(self._objects), {})

# ────────────────────────────────────────────────────────────────
# 2. Mock SubscribersHub
# ────────────────────────────────────────────────────────────────
class FakeStamp:
    def __init__(self, sec=0, nanosec=0):
        self.sec = sec; self.nanosec = nanosec

class FakeHeader:
    def __init__(self):
        self.stamp = FakeStamp()

class FakeOrigin:
    def __init__(self, x=0.0, y=0.0, z=0.0):
        self.position = type("P", (), {"x": x, "y": y, "z": z})()

class FakeMapInfo:
    def __init__(self, width, height, resolution):
        self.width = width
        self.height = height
        self.resolution = resolution
        self.origin = FakeOrigin()

class FakeOccupancyGrid:
    def __init__(self, width, height, resolution, data_list):
        self.header = FakeHeader()
        self.info = FakeMapInfo(width, height, resolution)
        self.data = data_list

class FakeHub:
    def __init__(self):
        self._slot = None
    def has(self, kind):
        return kind == "occupancy_grid" and self._slot is not None
    def latest(self, kind):
        return self._slot
    def inject(self, msg, count=1):
        """Inject an OccupancyGrid (simulating slot.write after ros2 topic pub)."""
        self._slot = (msg, 1700000000.0, count)

# ────────────────────────────────────────────────────────────────
# 3. Real _occupancy_bfs logic (copied verbatim from mcp_tools.py L109‑L152)
#    numpy replaced with pure Python equivalents, preserving all divisions / reshape semantics.
# ────────────────────────────────────────────────────────────────
_GOAL_NEAR_CLEARANCE_M = 0.4
_GOAL_NEAR_SEARCH_M = 6.0
_GOAL_NEAR_ROBOT_RADIUS_M = 0.3
_GOAL_NEAR_RING_STEP_M = 0.1
_GOAL_NEAR_ANGLE_STEPS = [0.0, 0.2, -0.2, 0.4, -0.4, 0.6, -0.6,
                          0.8, -0.8, 1.0, -1.0, 1.2, -1.2,
                          1.4, -1.4, 1.6, -1.6]

def fake_numpy_reshape_int8(data_bytes, w, h):
    """Simulate np.frombuffer(data, int8).reshape(h, w).

    Numpy semantics: if byte count != h*w, raise ValueError.
    """
    expected = h * w
    actual = len(data_bytes)
    if actual != expected:
        raise ValueError(
            f"cannot reshape array of size {actual} into shape ({h},{w})"
        )
    return [[0 for _ in range(w)] for _ in range(h)]  # dummy h x w array

def fake_numpy_blocked_any(grid, w, h):
    """Simulate (grid > 50).any() slice check: always returns False (free map).

    We want to trigger the prior divisions, not go deep into path search.
    """
    return False

def _occupancy_bfs_ORIGINAL(grid_msg, target_x, target_y, approach_ang):
    """mcp_tools.py L109‑L152 original logic, **without any fixes**."""
    info = grid_msg.info
    w, h = int(info.width), int(info.height)
    res = float(info.resolution)                         # ← L120: res may be 0
    ogx = float(info.origin.position.x)
    ogy = float(info.origin.position.y)
    # Secondary sink: L123
    grid = fake_numpy_reshape_int8(bytes(grid_msg.data), w, h)
    # blocked = grid > 50
    # infl = max(1, ceil((0.3 + 0.4) / res))             ← L129‑130: SINK #1 (division by zero)
    infl = max(1, int(math.ceil(
        (_GOAL_NEAR_ROBOT_RADIUS_M + _GOAL_NEAR_CLEARANCE_M) / res)))

    def is_safe(gx, gy):
        if gx - infl < 0 or gy - infl < 0 or gx + infl >= w or gy + infl >= h:
            return False
        return not fake_numpy_blocked_any(grid, w, h)

    standoff = 0.5
    n_rings = int(_GOAL_NEAR_SEARCH_M / _GOAL_NEAR_RING_STEP_M) + 1
    for i in range(n_rings):
        r = standoff + i * _GOAL_NEAR_RING_STEP_M
        for dth in _GOAL_NEAR_ANGLE_STEPS:
            ang = approach_ang + dth
            wx = target_x - math.cos(ang) * r
            wy = target_y - math.sin(ang) * r
            gx = int((wx - ogx) / res)                  # ← L147: SINK #2 (division by zero)
            gy = int((wy - ogy) / res)                  # ← L148: SINK #3 (division by zero)
            if 0 <= gx < w and 0 <= gy < h and is_safe(gx, gy):
                yaw = math.atan2(target_y - wy, target_x - wx)
                return wx, wy, yaw
    return None

# ────────────────────────────────────────────────────────────────
# 4. Trimmed goal_near top‑level logic (keeping all unfixed branches)
# ────────────────────────────────────────────────────────────────
def goal_near_UNFIXED(hub, registry, object_id):
    """Corresponds to mcp_tools.py goal_near, but without async and mcp decorator."""
    # In real code it awaits, but for component‑level sync is equivalent
    objs, _surfs = registry.snapshot()
    target = objs.get(object_id)
    if target is None:
        return {"reachable": False, "reason": f"unknown object_id '{object_id}'"}

    robot = next(
        (o for o in objs.values()
         if getattr(o, "is_robot", False) or str(o.cls).lower() == "robot"),
        None,
    )
    if robot is not None:
        approach_ang = math.atan2(
            float(target.pose.y) - float(robot.pose.y),
            float(target.pose.x) - float(robot.pose.x),
        )
    else:
        approach_ang = math.pi

    if hub is None or not hub.has("occupancy_grid"):
        return {"reachable": False, "reason": "no occupancy_grid available — mapping not running"}

    try:
        msg, _stamp, count = hub.latest("occupancy_grid")
    except Exception as e:
        return {"reachable": False, "reason": f"occupancy_grid hub error: {e}"}

    # L205: only checks width/height, not resolution or data length
    if msg is None or count == 0 or not msg.info.width or not msg.info.height:
        return {"reachable": False, "reason": "occupancy_grid empty"}

    # ↓ call the unfixed _occupancy_bfs
    found = _occupancy_bfs_ORIGINAL(msg, float(target.pose.x), float(target.pose.y), approach_ang)
    if found is None:
        return {"reachable": False, "reason": f"no free cell within {_GOAL_NEAR_SEARCH_M:.1f}m"}
    gx, gy, yaw = found
    return {"reachable": True, "x": gx, "y": gy, "yaw": yaw, "reason": "approach pose"}

# ────────────────────────────────────────────────────────────────
# 5. Build test scenarios and run
# ────────────────────────────────────────────────────────────────
def run_poc():
    # Scenario: target object at (4.5, 5.0), robot at (3.0, 3.0)
    potted_plant = FakeSceneObject(
        object_id="scene.object.potted_plant_001",
        cls="potted_plant",
        x=4.5, y=5.0, z=0.3,
    )
    robot = FakeSceneObject(
        object_id="scene.object.robot_001",
        cls="robot",
        x=3.0, y=3.0, z=0.0,
        is_robot=True,
    )
    registry = FakeRegistry([potted_plant, robot])

    # ── Case A: clean map (normal resolution) ───────────────────────
    print("=" * 60)
    print("Case A: clean occupancy_grid (resolution=0.05, width=10, height=10, data=[0]*100)")
    hub = FakeHub()
    normal_grid = FakeOccupancyGrid(
        width=10, height=10, resolution=0.05,
        data_list=[0] * 100,
    )
    hub.inject(normal_grid)
    try:
        r = goal_near_UNFIXED(hub, registry, "scene.object.potted_plant_001")
        print(f"  goal_near result: reachable={r['reachable']}, reason={r.get('reason','')}")
        print("  [INFO] clean case -> normal")
    except Exception as e:
        print(f"  [UNEXPECTED] clean case threw: {type(e).__name__}: {e}")

    # ── Case B: resolution=0 (main path for Issue #05) ─────────────────────
    print()
    print("=" * 60)
    print("Case B: poisoned occupancy_grid (resolution=0.0, width=4, height=4, data=[0]*16)")
    print("        ↑ data length 16 == 4*4, avoids the #04 reshape path")
    hub2 = FakeHub()
    poisoned_res0 = FakeOccupancyGrid(
        width=4, height=4, resolution=0.0,      # ← fatal: resolution=0
        data_list=[0] * 16,                      # ← but length is self‑consistent
    )
    hub2.inject(poisoned_res0)
    try:
        r = goal_near_UNFIXED(hub2, registry, "scene.object.potted_plant_001")
        print(f"  goal_near result: reachable={r['reachable']} — should not have reached here!")
        print("  [FAIL] resolution=0 did not raise exception, vulnerability not reproduced!")
        return 1
    except ZeroDivisionError as e:
        print(f"  [VULNERABILITY CONFIRMED] ZeroDivisionError: {e}")
        print("  ← matches real deployment: mcp tool error: ... float division by zero")
    except Exception as e:
        print(f"  [OTHER EXCEPTION] {type(e).__name__}: {e}")

    # ── Case C: data length mismatch (mirror of Issue #04 in MCP entry) ───────
    print()
    print("=" * 60)
    print("Case C: poisoned occupancy_grid (resolution=0.05, width=4, height=4, data=[0]*3)")
    print("        ↑ resolution normal, but data has only 3 elements (should be 16)")
    hub3 = FakeHub()
    poisoned_short = FakeOccupancyGrid(
        width=4, height=4, resolution=0.05,
        data_list=[0, 0, 0],                  # ← length insufficient
    )
    hub3.inject(poisoned_short)
    try:
        r = goal_near_UNFIXED(hub3, registry, "scene.object.potted_plant_001")
        print(f"  goal_near result: reachable={r['reachable']} — should not have reached here!")
        print("  [FAIL] data length mismatch did not raise exception, mirror path not reproduced!")
        return 1
    except ValueError as e:
        print(f"  [VULNERABILITY CONFIRMED] ValueError: {e}")
        print("  ← matches real deployment: mcp tool error ... cannot reshape array of size 3 into shape (4,4)")
    except Exception as e:
        print(f"  [OTHER EXCEPTION] {type(e).__name__}: {e}")

    # ── Case D: negative resolution (same root cause as #05) ─────────────────
    print()
    print("=" * 60)
    print("Case D: poisoned occupancy_grid (resolution=-0.05, width=4, height=4, data=[0]*16)")
    hub4 = FakeHub()
    poisoned_neg = FakeOccupancyGrid(
        width=4, height=4, resolution=-0.05,    # negative resolution
        data_list=[0] * 16,
    )
    hub4.inject(poisoned_neg)
    try:
        r = goal_near_UNFIXED(hub4, registry, "scene.object.potted_plant_001")
        # Negative resolution does not raise division by zero, but produces negative/out‑of‑bounds grid coords,
        # which may cause is_safe to incorrectly reject.
        print(f"  goal_near result: reachable={r.get('reachable','?')}, reason={r.get('reason','')}")
        print("  [INFO] resolution<0 does not divide by zero but yields wrong semantic coordinates (gx/gy negative)")
    except Exception as e:
        print(f"  [SIDE EFFECT] {type(e).__name__}: {e}")

    print()
    print("=" * 60)
    print("Component‑level PoC summary:")
    print("  Case A (clean): normal ✓")
    print("  Case B (res=0): ZeroDivisionError — Issue #05 confirmed ✓")
    print("  Case C (len<):  ValueError reshape — Issue #04 mirror path confirmed ✓")
    print("  Case D (res<0): coordinate semantic error ✓")
    return 0

if __name__ == "__main__":
    sys.exit(run_poc())
```

---

### 4.1 Design Principles

1. **Validation at entry**: Before entering `_occupancy_bfs`, centrally validate OccupancyGrid metadata **numeric validity** and **structural consistency** inside `goal_near`.
2. **Defence in depth**: Add another validation layer at the start of `_occupancy_bfs` (double‑checking, in case future callers bypass the outer check).
3. **Structured failure**: On validation failure, `goal_near` returns `reachable=False` with an explicit `reason`, rather than throwing an implementation exception up to the MCP layer.
4. **Observability**: Each time an invalid map is discarded, log a warning with specific field values to help upstream providers debug.

### 4.2 Specific Code Changes

**File**: `system/scene/scene_service/mcp_tools.py`

#### 4.2.1 Add OccupancyGrid validation function

Insert before the definition of `_occupancy_bfs` (around L108) in `mcp_tools.py`:

```python
def _validate_occupancy_grid(msg) -> tuple[bool, str]:
    """Validate an OccupancyGrid before use in BFS or rendering.

    Returns (ok: bool, reason: str).  reason is empty when ok=True.
    """
    if msg is None:
        return False, "occupancy_grid message is None"

    info = msg.info
    w, h = int(info.width), int(info.height)
    res = float(info.resolution)

    # 1) dimensions must be positive
    if w <= 0 or h <= 0:
        return False, f"invalid dimensions width={w} height={h}"

    # 2) resolution must be positive (exclude 0, negative, NaN, Inf)
    try:
        # math.isfinite catches inf and nan; then also require > 0
        if not math.isfinite(res) or res <= 0.0:
            return False, (
                f"invalid resolution={res!r} — "
                "must be finite and strictly positive"
            )
    except (TypeError, ValueError):
        return False, f"non-numeric resolution={info.resolution!r}"

    # 3) data length must equal width * height (prevents Issue #04 from triggering via MCP entry)
    expected = w * h
    actual = len(msg.data)
    if actual != expected:
        return False, (
            f"data length mismatch: len(data)={actual} != "
            f"width*height={w}*{h}={expected}"
        )

    # 4) Soft sanity check on resolution: 0.001m (1mm) ~ 5.0m – warn but don't reject
    #    (keeping compatibility with very coarse or very fine grids)
    if res < 0.001 or res > 5.0:
        log.warning(
            "[scene/goal_near] occupancy_grid resolution=%.6f m is outside "
            "typical range [0.001, 5.0] — continuing anyway",
            res,
        )

    return True, ""
```

#### 4.2.2 Add defensive validation inside `_occupancy_bfs`

Insert a validation check at the beginning of `_occupancy_bfs` (around original L118‑L120), replacing the unvalidated flow. The function start becomes:

```python
def _occupancy_bfs(
    grid_msg, target_x: float, target_y: float, approach_ang: float
) -> tuple[float, float, float] | None:
    """Sweep rings of free cells around (target_x, target_y) on the
    occupancy grid; pick the first one with enough clearance and face
    back at the target. Returns (x, y, yaw) or None if nothing free.

    Raises ValueError iff the grid is structurally invalid and the
    caller forgot to run _validate_occupancy_grid first.
    """
    import numpy as np

    # Defence‑in‑depth: even if caller didn't validate, do it here
    ok, bad_reason = _validate_occupancy_grid(grid_msg)
    if not ok:
        log.warning("[scene/_occupancy_bfs] invalid grid, aborting BFS: %s", bad_reason)
        return None

    info = grid_msg.info
    w, h = int(info.width), int(info.height)
    res = float(info.resolution)  # safe — validated above (res > 0 and finite)
    ogx = float(info.origin.position.x)
    ogy = float(info.origin.position.y)
    grid = np.frombuffer(bytes(grid_msg.data), dtype=np.int8).reshape(h, w)
    # ... (rest of original code, now without division‑by‑zero risk)
```

#### 4.2.3 Validate at the `goal_near` top level and return structured failure

In the original `goal_near`, replace the block from the `try/except` around L198 up to the call to `_occupancy_bfs`. Change from:

```python
    try:
        msg, _stamp, count = _HUB.latest("occupancy_grid")
    except Exception as e:  # noqa: BLE001
        return GoalNear_Response(
            reachable=False, x=0.0, y=0.0, yaw=0.0,
            reason=f"occupancy_grid hub error: {e}",
        )
    if msg is None or count == 0 or not msg.info.width or not msg.info.height:
        return GoalNear_Response(
            reachable=False, x=0.0, y=0.0, yaw=0.0,
            reason="occupancy_grid empty — wait for mapping to publish",
        )

    found = _occupancy_bfs(
        msg,
        float(target.pose.x),
        float(target.pose.y),
        approach_ang,
    )
```

to:

```python
    try:
        msg, _stamp, count = _HUB.latest("occupancy_grid")
    except Exception as e:  # noqa: BLE001
        return GoalNear_Response(
            reachable=False, x=0.0, y=0.0, yaw=0.0,
            reason=f"occupancy_grid hub error: {e}",
        )
    if msg is None or count == 0 or not msg.info.width or not msg.info.height:
        return GoalNear_Response(
            reachable=False, x=0.0, y=0.0, yaw=0.0,
            reason="occupancy_grid empty — wait for mapping to publish",
        )

    # FIX: validate resolution legality + data length consistency
    grid_ok, grid_reason = _validate_occupancy_grid(msg)
    if not grid_ok:
        log.warning(
            "[scene/goal_near] discarding invalid occupancy_grid: %s; "
            "returning reachable=false to caller",
            grid_reason,
        )
        return GoalNear_Response(
            reachable=False, x=0.0, y=0.0, yaw=0.0,
            reason=(
                f"invalid occupancy_grid metadata ({grid_reason}); "
                "map upstream likely publishing bad data — "
                "try again once mapping recovers"
            ),
        )

    found = _occupancy_bfs(
        msg,
        float(target.pose.x),
        float(target.pose.y),
        approach_ang,
    )
```

#### 4.2.4 Top‑level fallback: wrap `goal_near` body in try/except (last line of defence)

Wrap the entire body of `goal_near` (after the existing `_REGISTRY is None` check) in a `try/except` to ensure that any unexpected implementation exception **no longer** leaks as an MCP `Error executing tool ...` text, but instead returns a controlled `reachable=False`:

Within `goal_near`, keep the existing `_REGISTRY is None` check, then wrap the rest of the logic in:

```python
@mcp_contract(mcp, contract_id="robonix/system/scene/goal_near")
async def goal_near(req: GoalNear_Request) -> GoalNear_Response:
    if _REGISTRY is None:
        raise RuntimeError("scene mcp_tools.attach_state was never called")
    try:
        # ── All original logic (from objs, _surfs = await _REGISTRY.snapshot()
        #     down to the final return GoalNear_Response(...)) goes here ──
        ...
    except Exception as exc:  # noqa: BLE001 — last‑resort catch
        log.exception("[scene/goal_near] unexpected failure: object_id=%s", req.object_id)
        return GoalNear_Response(
            reachable=False, x=0.0, y=0.0, yaw=0.0,
            reason=f"goal_near internal error: {type(exc).__name__}",
        )
```

(Inside the block, existing `return GoalNear_Response(...)` statements remain – they will return normally and not hit the outer except.)

### 4.3 Summary of Changes

| File | Location | Change |
|------|----------|--------|
| `mcp_tools.py` | around L108 | Add `_validate_occupancy_grid(msg)` that validates resolution legality + data length consistency |
| `mcp_tools.py` | start of `_occupancy_bfs` (L116) | Add defence‑in‑depth validation call; on failure `log.warning` + return None |
| `mcp_tools.py` | `goal_near` L198‑L210 area | Before BFS, call `_validate_occupancy_grid`; on failure return structured `reachable=false` + reason |
| `mcp_tools.py` | `goal_near` body | Wrap whole body in `try/except Exception` so that exceptions no longer leak to MCP layer |

### 4.4 Behaviour After Fix

| Scenario | Before Fix | After Fix |
|----------|------------|-----------|
| Clean map resolution=0.05 | `reachable=true/false` + normal reason | Behaviour unchanged ✓ |
| resolution=0.0 (Case B) | `ZeroDivisionError` → MCP error text `float division by zero` → caller crashes | `reachable=false` + `reason="invalid occupancy_grid metadata (invalid resolution=0.0 ...)"` + log.warning |
| data length mismatch (Case C) | `ValueError: cannot reshape ...` → MCP error text | `reachable=false` + `reason="invalid occupancy_grid metadata (data length mismatch ...)"` + log.warning |
| negative resolution (Case D) | silently produces wrong semantic coordinates (gx,gy negative), BFS results unreliable | Same as resolution=0, structured rejection |
| resolution=NaN/Inf | `ZeroDivisionError` or nonsense coordinates | Same as resolution=0 |
| Other unexpected exceptions | Leak as `Error executing tool goal_near: <exc>` | Caught + log.exception + `reachable=false, reason="goal_near internal error: <ExcType>"` |

---

## 5. Impact Assessment

| Dimension | Assessment |
|-----------|------------|
| **Vulnerability type** | Upstream state pollution + missing metadata boundary validation → capability‑level denial of service |
| **Primary crash** | `ZeroDivisionError: float division by zero` (3 division sites) |
| **Secondary crash** | `ValueError: cannot reshape ...` (mirror of Issue #04 in MCP entry) |
| **Prerequisite for attack** | Ability to publish to ROS2 `/map` (or any occupancy_grid topic subscribed by scene) |
| **Cost of attack** | One `ros2 topic pub --once` command, or any equivalent publisher |
| **MCP request body** | Completely legitimate (passes JSON‑Schema validation) – no malformed payload needed |
| **Impact scope** | The `robonix/system/scene/goal_near` capability becomes unusable until the slot is overwritten by a valid map |
| **Effect on Pilot planning** | A `goal_near` failure typically degrades to abandoning the approach, but the "mcp tool error" text may cause the planner to misinterpret the failure as a toolchain issue and retry/alert, rather than interpreting it as "object unreachable" |
| **CVSS v4 estimate** | AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:L → **Medium (3.1‑5.9)** |

---

### 6.1 Deployment‑level regression verification (after fix, with pollution)

```bash
# 1) Deploy the fixed scene version, run baseline
source /opt/ros/humble/setup.bash
export ROBONIX_HOME=/tmp/robonix-home-audit
export RMW_IMPLEMENTATION=rmw_zenoh_cpp
cd /home/ysq/robonix/examples/webots
rbnx ask "Call scene goal_near for object_id=scene.object.potted_plant_001, print raw result"
# Expected: reachable=... structured result, no ZeroDivisionError

# 2) Pollute again (same command as in PoC)
docker exec robonix_tiago_sim bash -lc \
  'source /opt/ros/humble/setup.bash && \
   ros2 topic pub --once /map nav_msgs/msg/OccupancyGrid \
   "{header: {frame_id: map}, info: {resolution: 0.0, width: 4, height: 4, origin: {position: {x: 0.0, y: 0.0, z: 0.0}, orientation: {w: 1.0}}}, data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}"'

# 3) Same normal goal_near call
rbnx ask "Call scene goal_near for object_id=scene.object.potted_plant_001, print raw result"
# Expected (after fix):
#   reachable=false
#   reason contains "invalid occupancy_grid metadata" + "invalid resolution=0.0"
#   No ZeroDivisionError, no "Error executing tool goal_near"

# 4) Check scene logs for the warning
rbnx logs -t scene --tail 50 | grep "invalid occupancy_grid metadata"
# Expected: one warning entry

# 5) The data length mismatch mirror path should also return structured failure
docker exec robonix_tiago_sim bash -lc \
  'source /opt/ros/humble/setup.bash && \
   ros2 topic pub --once /map nav_msgs/msg/OccupancyGrid \
   "{header: {frame_id: map}, info: {resolution: 0.05, width: 4, height: 4, origin: {position: {x: 0.0, y: 0.0, z: 0.0}, orientation: {w: 1.0}}}, data: [0, 0, 0]}"'

rbnx ask "Call scene goal_near for object_id=scene.object.potted_plant_001, print raw result"
# Expected (after fix):
#   reachable=false
#   reason contains "data length mismatch"
#   No ValueError, no "Error executing tool goal_near"
```

### 6.2 Component‑level unit test suggestions (Python pytest)

Add `test_goal_near_grid_validation.py` under `system/scene/tests/` covering:

| Test case | Input | Expected |
|-----------|-------|----------|
| `test_clean_grid_passes` | w=10, h=10, res=0.05, data=[0]*100 | BFS returns normally or None, no exception |
| `test_resolution_zero` | w=4, h=4, res=0.0, data=[0]*16 | goal_near → `reachable=false`, reason contains `invalid resolution`; no ZeroDivisionError |
| `test_resolution_negative` | w=4, h=4, res=-0.05, data=[0]*16 | goal_near → `reachable=false`; does not use wrong coordinates |
| `test_resolution_nan` | w=4, h=4, res=nan, data=[0]*16 | goal_near → `reachable=false` |
| `test_resolution_inf` | w=4, h=4, res=inf, data=[0]*16 | goal_near → `reachable=false` |
| `test_data_too_short` | w=4, h=4, res=0.05, data=[0]*3 | goal_near → `reachable=false`, reason contains `data length mismatch`; no ValueError |
| `test_data_too_long` | w=4, h=4, res=0.05, data=[0]*100 | goal_near → `reachable=false` |
| `test_width_zero` | w=0, h=4, res=0.05, data=[0]*0 | goal_near → `reachable=false` (existing branch + new validation double‑hit) |
| `test_last_resort_catch` | Force `hub.latest()` to raise RuntimeError | goal_near → `reachable=false`, reason contains `internal error`; does not leak to MCP layer |

enkerewpo:
Fixes #226. Fixes #227. Relates to #229.

## What was wrong

ROS 2 does not check that a message is self-consistent. An `OccupancyGrid` whose `data` length disagrees with `width * height`, or an `Image` whose buffer does not match its declared size and encoding, is type-valid and is delivered to every subscriber. Scene then reshaped the buffer, and `reshape` raises.

The raising happened inside request handlers, which is what made it matter. The debug UI answered 500 on `/api/state` and `/api/camera` for as long as the malformed message stayed the latest one, so anyone able to publish on `/map` could hold the interface down. The same contamination reached `goal_near` through the MCP path, where a resolution of zero also divided by zero.

Separately, that UI has no authentication and its annotation endpoints read and write map data, and it defaulted to binding every interface.

## What changed

**One rule, stated once.** `scene_service/message_shape.py` says what length a grid or an image of a given size and encoding must carry. It is pure and imports nothing. Both decode sites ask it, so the rule cannot be applied at one and forgotten at the other, and it can be tested without a ROS or robonix runtime.

**A malformed frame is dropped, not raised.** The rest of the page keeps working and the next well-formed message recovers the view, with one warning naming what was discarded.

**The planner half was already fixed.** The `goal_planner` refactor rejects a non-positive resolution and a mismatched cell count, so #227 no longer reproduces. Nothing tested either check, so nothing would have noticed them going away again — that is what the new tests are for.

**The UI defaults to loopback.** All three places that choose the bind address — the resolver and both launchers — now default to `127.0.0.1`. A launcher still passing `0.0.0.0` would have made the library default meaningless, so a test checks that too. The container runs with `--network host`, so the service's bind is the host's bind and nothing about port publishing changes. `SCENE_WEB_HOST=0.0.0.0`, or `web_host` in the Scene config, opens it again in one line, and the service logs a warning whenever it is bound wide so the choice stays visible.

## Tests

`system/scene/tests/test_malformed_grid_inputs.py`, 21 cases: the rule itself, both planner entry points, and every encoding Scene decodes.

These and `test_goal_planner.py` now run in the pull-request CI job. They import neither `robonix_api` nor the generated MCP types, so the lightweight dependency set already there is enough. Nothing on a pull request exercised these guards before, which is how they came to be missing in the first place.

Locally: 201 passed against a 200 baseline, with the same 11 pre-existing failures on both sides — all of them `ModuleNotFoundError` for workspace and codegen artifacts that are absent outside a built package.

## Note on #229

Authentication itself is not added here. The endpoints remain open to whoever can reach the port; what changed is that reaching it now takes a deliberate act. Whether this UI should carry real authentication is a larger question than a default, and worth deciding separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,17 @@ jobs:
           system/scene/tests/test_perception_vlm_geometry.py
           system/scene/tests/test_image_relations.py
 
+      # The goal planner and the message-shape rule are pure and import neither
+      # robonix_api nor the generated MCP types, so they run under the same
+      # lightweight dependency set. Without them here nothing on a pull request
+      # exercises the guards against a malformed occupancy grid, which is how
+      # those guards came to be missing in the first place.
+      - name: Test Scene goal planning and upstream message validation
+        run: >-
+          python3 -m pytest -q
+          system/scene/tests/test_goal_planner.py
+          system/scene/tests/test_malformed_grid_inputs.py
+
   check:
     name: Rust (fmt + clippy + build + tests)
     runs-on: ubuntu-22.04
--- a/system/scene/scene_service/goal_planner.py
+++ b/system/scene/scene_service/goal_planner.py
@@ -7,6 +7,7 @@ import math
 from collections.abc import Iterable, Sequence
 
 from .geometry import point_in_polygon, polygon_centroid
+from .message_shape import occupancy_grid_is_well_formed
 from .robot_geometry import RobotFootprint
 
 Point = tuple[float, float]
@@ -157,7 +158,7 @@ def _grid_array(grid_msg, *, width: int, height: int):
         grid = np.frombuffer(data, dtype=np.int8)
     else:
         grid = np.asarray(data, dtype=np.int8)
-    if grid.size != width * height:
+    if not occupancy_grid_is_well_formed(width, height, grid.size):
         raise ValueError(
             f"occupancy grid contains {grid.size} cells; expected {width * height}"
         )
--- a/system/scene/scene_service/message_shape.py
+++ b/system/scene/scene_service/message_shape.py
@@ -1,0 +1,81 @@
+# SPDX-License-Identifier: MulanPSL-2.0
+"""What makes an upstream image or occupancy grid well formed.
+
+ROS 2 does not check that a message is self-consistent. An `OccupancyGrid`
+whose `data` is shorter than `width * height`, or an `Image` whose buffer does
+not match its declared size and encoding, is type-valid and will be delivered
+to every subscriber. Both then reach code that reshapes the buffer, and
+`reshape` raises.
+
+That mattered because the raising happened inside request handlers: the Scene
+debug UI answered 500 on `/api/state` and `/api/camera` for as long as the
+malformed message stayed the latest one, and the `goal_near` capability failed
+the same way through the MCP path. Anyone able to publish on `/map` could hold
+the interface down.
+
+The rule belongs in one place rather than at each decode site, so it cannot be
+applied at one and forgotten at another. These functions are pure and import
+nothing, so they can be tested without a ROS or robonix runtime.
+
+Reported as issues #226 and #227.
+"""
+
+from __future__ import annotations
+
+# Bytes per pixel for the encodings Scene decodes. An encoding that is absent
+# is one Scene does not render, and its buffer length is not this module's
+# business.
+_BYTES_PER_PIXEL = {
+    "rgb8": 3,
+    "bgr8": 3,
+    "rgba8": 4,
+    "bgra8": 4,
+    "mono8": 1,
+    "32fc1": 4,
+    "16uc1": 2,
+}
+
+
+def occupancy_cells_expected(width: int, height: int) -> int:
+    """How many cells a grid of this size must carry."""
+    return int(width) * int(height)
+
+
+def occupancy_grid_is_well_formed(width: int, height: int, data_len: int) -> bool:
+    """Whether a grid can be reshaped to its declared size.
+
+    A zero dimension is rejected too: it is not decodable, and treating it as
+    "empty but fine" pushes an empty array into code that assumes cells exist.
+    """
+    if int(width) <= 0 or int(height) <= 0:
+        return False
+    return int(data_len) == occupancy_cells_expected(width, height)
+
+
+def image_bytes_expected(width: int, height: int, encoding: str) -> int | None:
+    """How many bytes an image of this size and encoding must carry.
+
+    `None` for an encoding Scene does not decode — the caller rejects those on
+    their own grounds, and guessing a size for them would be inventing one.
+    """
+    per_pixel = _BYTES_PER_PIXEL.get((encoding or "").lower())
+    if per_pixel is None:
+        return None
+    return int(width) * int(height) * per_pixel
+
+
+def image_is_well_formed(
+    width: int, height: int, encoding: str, data_len: int
+) -> bool:
+    """Whether an image buffer matches its declared size and encoding.
+
+    An unsupported encoding returns True: this function answers "is the buffer
+    the right length", and for an encoding with no defined length that question
+    does not apply. The decoder still rejects it.
+    """
+    if int(width) <= 0 or int(height) <= 0:
+        return False
+    expected = image_bytes_expected(width, height, encoding)
+    if expected is None:
+        return True
+    return int(data_len) == expected
--- a/system/scene/scene_service/service.py
+++ b/system/scene/scene_service/service.py
@@ -1791,6 +1791,19 @@ async def _run_active(config: dict) -> None:
         web_server = uvicorn.Server(web_uv)
         web_task = asyncio.create_task(web_server.serve(), name="scene-web-http")
         log.info("web UI on http://%s:%d", web_host, web_port)
+        # This surface carries no authentication — it is a debug and operator
+        # UI, and its annotation endpoints read and write map data. That is
+        # tenable on loopback and is a decision on any other address, so an
+        # operator who did not make that decision deliberately should be told
+        # they have. `web_host: 127.0.0.1` in the Scene config, or
+        # SCENE_WEB_HOST, keeps it local.
+        if web_host not in ("127.0.0.1", "::1", "localhost"):
+            log.warning(
+                "web UI is bound to %s and has no authentication: anyone who "
+                "can reach %s:%d may read and modify map annotations. Set "
+                "web_host: 127.0.0.1 (or SCENE_WEB_HOST) to keep it local.",
+                web_host, web_host, web_port,
+            )
 
     log.info(
         "scene up; cap=%s mcp=%s observations=%d",
--- a/system/scene/scene_service/service.py
+++ b/system/scene/scene_service/service.py
@@ -1758,8 +1758,8 @@ async def _run_active(config: dict) -> None:
     )
     mcp_tools.attach_object_mutations(object_mutations)
 
-    # SCENE_WEB_HOST are environment fallbacks; an explicit Scene config file
-    # can set web_host: 127.0.0.1 to keep this operator surface local-only.
+    # SCENE_WEB_HOST is the environment fallback; an explicit Scene config file
+    # can set web_host. Both default to loopback — see web_binding.
     web_port = int(
         int(config.get("web_port") or "0")
         if config.get("web_port") is not None and config.get("web_port") != ""
@@ -1791,17 +1791,15 @@ async def _run_active(config: dict) -> None:
         web_server = uvicorn.Server(web_uv)
         web_task = asyncio.create_task(web_server.serve(), name="scene-web-http")
         log.info("web UI on http://%s:%d", web_host, web_port)
-        # This surface carries no authentication — it is a debug and operator
-        # UI, and its annotation endpoints read and write map data. That is
-        # tenable on loopback and is a decision on any other address, so an
-        # operator who did not make that decision deliberately should be told
-        # they have. `web_host: 127.0.0.1` in the Scene config, or
-        # SCENE_WEB_HOST, keeps it local.
+        # Loopback is the default. Anything else was asked for, and is worth
+        # one line in the log because this surface has no authentication and
+        # its annotation endpoints write map data — whoever reaches it can
+        # change what the robot believes about its world.
         if web_host not in ("127.0.0.1", "::1", "localhost"):
             log.warning(
-                "web UI is bound to %s and has no authentication: anyone who "
-                "can reach %s:%d may read and modify map annotations. Set "
-                "web_host: 127.0.0.1 (or SCENE_WEB_HOST) to keep it local.",
+                "web UI is bound to %s, not loopback, and has no "
+                "authentication: anyone who can reach %s:%d may read and "
+                "modify map annotations.",
                 web_host, web_host, web_port,
             )
 
--- a/system/scene/scene_service/web.py
+++ b/system/scene/scene_service/web.py
@@ -32,6 +32,8 @@ from starlette.routing import Route
 
 from robonix_api import ATLAS
 
+from .message_shape import image_is_well_formed, occupancy_grid_is_well_formed
+
 from .annotations import validate_annotation_fields
 from .map_binding import sanitize_map_id as _sanitize_map_id
 from .map_meta import make_meta
@@ -325,7 +327,18 @@ def _occupancy_payload(hub: Any) -> Optional[dict]:
         return None
     info = msg.info
     w, h = int(info.width), int(info.height)
-    if w == 0 or h == 0:
+    # A publisher is not required to be honest. `data` is sized by whoever
+    # published it, and a grid whose length disagrees with width * height is
+    # type-valid ROS 2 — nothing upstream rejects it. Reshaping it raises, and
+    # an exception here becomes a 500 on /api/state for as long as that message
+    # stays the latest one, which takes the whole debug UI down. Dropping the
+    # frame keeps the rest of the page working and lets the next well-formed
+    # grid recover it.
+    if not occupancy_grid_is_well_formed(w, h, len(msg.data)):
+        log.warning(
+            "occupancy_grid dropped: %d cells for a %dx%d grid",
+            len(msg.data), w, h,
+        )
         return None
     # nav_msgs/OccupancyGrid data is row-major bottom-up int8 in
     # [-1, 100]: -1 unknown, 0 free, 100 occupied. Render as grayscale:
@@ -366,9 +379,16 @@ def _image_to_png_b64(msg: Any, *, kind: str) -> Optional[dict]:
     except ImportError:
         return None
     h, w = int(msg.height), int(msg.width)
-    if h == 0 or w == 0:
-        return None
     enc = (msg.encoding or "").lower()
+    # Same reasoning as the occupancy grid above: the buffer is sized by the
+    # publisher, and one that disagrees with the declared size makes reshape
+    # raise, which surfaces as a 500 on /api/camera.
+    if not image_is_well_formed(w, h, enc, len(msg.data)):
+        log.warning(
+            "%s image dropped: %d bytes for a %dx%d %s frame",
+            kind, len(msg.data), w, h, enc,
+        )
+        return None
     arr: Any = None
     out_mode = "RGB"
     if kind == "rgb":
--- a/system/scene/scene_service/web_binding.py
+++ b/system/scene/scene_service/web_binding.py
@@ -1,9 +1,14 @@
 # SPDX-License-Identifier: MulanPSL-2.0
 """Resolve the Scene debug UI bind address.
 
-The port is already deployment-configurable.  Keep the historical all-interface
-default for existing deployments, while allowing safety-sensitive robot
-profiles to bind the UI to loopback explicitly.
+The UI carries no authentication and its annotation endpoints read and write
+map data, so reaching it is enough to alter what the robot believes about its
+world. The default is therefore loopback: exposing that surface to a network
+is a decision, and a decision should be made rather than inherited.
+
+`web_host` in the Scene config, or SCENE_WEB_HOST, opens it deliberately —
+a deployment whose operator UI is reached from another machine sets one of
+them, and the service says at startup when it is bound wide.
 """
 
 from __future__ import annotations
@@ -24,7 +29,7 @@ def resolve_web_host(
     raw = (
         config["web_host"]
         if "web_host" in config
-        else env.get("SCENE_WEB_HOST", "0.0.0.0")
+        else env.get("SCENE_WEB_HOST", "127.0.0.1")
     )
     if not isinstance(raw, str):
         raise ValueError("Scene web_host/SCENE_WEB_HOST must be a string")
--- a/system/scene/scripts/start.sh
+++ b/system/scene/scripts/start.sh
@@ -168,6 +168,10 @@ if [[ "${ROBONIX_FORCE_CPU:-0}" != "1" ]]; then
     fi
 fi
 
+# SCENE_WEB_HOST below defaults to loopback. The container runs with
+# --network host, so what the service binds is what the host exposes, and the
+# web UI has no authentication while its annotation endpoints write map data.
+# Set SCENE_WEB_HOST=0.0.0.0 to reach it from another machine.
 exec docker run --rm \
     --name "$CT" \
     --entrypoint /scene/docker/entrypoint.sh \
@@ -180,7 +184,7 @@ exec docker run --rm \
     -e ROBONIX_CAPABILITY_ID="${ROBONIX_CAPABILITY_ID:-com.robonix.system.scene}" \
     -e ROBONIX_PKG_HOST_DIR="$(pwd)" \
     -e SCENE_WEB_PORT="${SCENE_WEB_PORT:-50107}" \
-    -e SCENE_WEB_HOST="${SCENE_WEB_HOST-0.0.0.0}" \
+    -e SCENE_WEB_HOST="${SCENE_WEB_HOST-127.0.0.1}" \
     -e SCENE_LOG_LEVEL="${SCENE_LOG_LEVEL:-INFO}" \
     -e SCENE_CG_FORCE_CPU="${SCENE_CG_FORCE_CPU:-}" \
     -e SCENE_CG_OBJ_MIN_POINTS="${SCENE_CG_OBJ_MIN_POINTS:-}" \
--- a/system/scene/scripts/start_native.sh
+++ b/system/scene/scripts/start_native.sh
@@ -66,7 +66,10 @@ mkdir -p "$SCENE_ANNOTATIONS_DIR"
 
 # Service knobs (same defaults the docker run wires via -e).
 export SCENE_WEB_PORT="${SCENE_WEB_PORT:-50107}"
-export SCENE_WEB_HOST="${SCENE_WEB_HOST-0.0.0.0}"
+# Loopback by default: the web UI has no authentication and its annotation
+# endpoints write map data. Set SCENE_WEB_HOST=0.0.0.0 to reach it from another
+# machine.
+export SCENE_WEB_HOST="${SCENE_WEB_HOST-127.0.0.1}"
 export SCENE_LOG_LEVEL="${SCENE_LOG_LEVEL:-INFO}"
 
 echo "[scene/native] python=$PY ros=$ROS_DISTRO rmw=${RMW_IMPLEMENTATION:-<default>} web=$SCENE_WEB_HOST:$SCENE_WEB_PORT"
--- a/system/scene/tests/conftest.py
+++ b/system/scene/tests/conftest.py
@@ -1,0 +1,14 @@
+# SPDX-License-Identifier: MulanPSL-2.0
+"""Put the Scene package root on sys.path for every test in this directory.
+
+pytest prepends the directory that holds a test file, not the package root, so
+`from scene_service import ...` only resolves when pytest is run from
+system/scene. The older tests each carry their own sys.path.insert; the ones
+that do not import nothing when CI runs pytest from the repository root, which
+is how test_goal_planner.py could sit in the tree unexercised. Doing it once
+here means a new test does not have to remember.
+"""
+import os
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
--- a/system/scene/tests/test_malformed_grid_inputs.py
+++ b/system/scene/tests/test_malformed_grid_inputs.py
@@ -1,0 +1,111 @@
+# SPDX-License-Identifier: MulanPSL-2.0
+"""Scene must survive a structurally inconsistent upstream message.
+
+Nothing in ROS 2 rejects an `OccupancyGrid` whose `data` length disagrees with
+`width * height`, or an `Image` whose buffer does not match its declared size
+and encoding. Both are type-valid and are delivered to every subscriber. Both
+used to reach code that reshaped the buffer, and `reshape` raises.
+
+The raising happened inside request handlers, which is what made it matter: the
+debug UI answered 500 on `/api/state` and `/api/camera` for as long as the bad
+message stayed the latest one, and `goal_near` failed the same way through the
+MCP path. Anyone able to publish on `/map` could hold the interface down.
+
+Both entry points now ask `message_shape` the same question, so these tests
+cover the rule itself and the fact that each caller applies it.
+
+Reported as issues #226 and #227.
+"""
+
+from types import SimpleNamespace
+
+import numpy as np
+import pytest
+
+from scene_service import goal_planner
+from scene_service.message_shape import (
+    image_bytes_expected,
+    image_is_well_formed,
+    occupancy_grid_is_well_formed,
+)
+
+
+def _grid(*, width=8, height=6, resolution=0.05, cells=None):
+    """An OccupancyGrid whose `data` length is independent of its declared
+    size, which is the whole point of these tests."""
+    if cells is None:
+        cells = width * height
+    return SimpleNamespace(
+        header=SimpleNamespace(frame_id="map"),
+        info=SimpleNamespace(
+            width=width,
+            height=height,
+            resolution=resolution,
+            origin=SimpleNamespace(position=SimpleNamespace(x=0.0, y=0.0)),
+        ),
+        data=np.zeros(cells, dtype=np.int8).tobytes(),
+    )
+
+
+# --- the rule ---------------------------------------------------------------
+
+
+def test_a_grid_is_well_formed_only_at_its_declared_size():
+    assert occupancy_grid_is_well_formed(8, 6, 48)
+    assert not occupancy_grid_is_well_formed(8, 6, 47)
+    assert not occupancy_grid_is_well_formed(8, 6, 49)
+
+
+@pytest.mark.parametrize(("width", "height"), [(0, 6), (8, 0), (-1, 6)])
+def test_a_grid_without_extent_is_not_well_formed(width, height):
+    assert not occupancy_grid_is_well_formed(width, height, 0)
+
+
+@pytest.mark.parametrize(
+    ("encoding", "per_pixel"),
+    [("rgb8", 3), ("bgr8", 3), ("rgba8", 4), ("bgra8", 4), ("mono8", 1),
+     ("32fc1", 4), ("16uc1", 2)],
+)
+def test_each_decoded_encoding_states_its_own_length(encoding, per_pixel):
+    assert image_bytes_expected(4, 3, encoding) == 4 * 3 * per_pixel
+    assert image_is_well_formed(4, 3, encoding, 4 * 3 * per_pixel)
+    assert not image_is_well_formed(4, 3, encoding, 4 * 3 * per_pixel - 1)
+
+
+def test_an_encoding_scene_does_not_decode_has_no_length_to_check():
+    # The decoder rejects it on its own grounds; inventing a size for it here
+    # would be guessing.
+    assert image_bytes_expected(4, 3, "yuv422") is None
+    assert image_is_well_formed(4, 3, "yuv422", 7)
+
+
+# --- the planner path (#227) ------------------------------------------------
+
+
+def test_zero_resolution_is_rejected_rather_than_dividing_by_zero():
+    with pytest.raises(ValueError, match="resolution"):
+        goal_planner._grid_metadata(_grid(resolution=0.0))
+
+
+@pytest.mark.parametrize("resolution", [-0.05, float("nan"), float("inf")])
+def test_a_resolution_that_cannot_scale_anything_is_rejected(resolution):
+    with pytest.raises(ValueError, match="resolution"):
+        goal_planner._grid_metadata(_grid(resolution=resolution))
+
+
+@pytest.mark.parametrize(("width", "height"), [(0, 6), (8, 0)])
+def test_an_empty_dimension_is_rejected(width, height):
+    with pytest.raises(ValueError):
+        goal_planner._grid_metadata(_grid(width=width, height=height))
+
+
+@pytest.mark.parametrize("cells", [10, 100])
+def test_a_mismatched_buffer_is_rejected_before_reshape(cells):
+    with pytest.raises(ValueError, match="cells"):
+        goal_planner._grid_array(_grid(width=8, height=6, cells=cells),
+                                 width=8, height=6)
+
+
+def test_a_well_formed_grid_still_decodes():
+    grid = _grid(width=8, height=6)
+    assert goal_planner._grid_array(grid, width=8, height=6).shape == (6, 8)
--- a/system/scene/tests/test_web_binding.py
+++ b/system/scene/tests/test_web_binding.py
@@ -19,11 +19,14 @@ def test_explicit_config_host_wins_over_environment():
 
 
 def test_environment_host_is_supported_for_launcher_compatibility():
-    assert resolve_web_host({}, {"SCENE_WEB_HOST": "127.0.0.1"}) == "127.0.0.1"
+    assert resolve_web_host({}, {"SCENE_WEB_HOST": "0.0.0.0"}) == "0.0.0.0"
 
 
-def test_existing_deployments_keep_legacy_default():
-    assert resolve_web_host({}, {}) == "0.0.0.0"
+def test_the_ui_is_local_unless_someone_opens_it():
+    # The UI has no authentication and its annotation endpoints write map
+    # data, so the default is loopback. Exposing it is a decision; a
+    # deployment that wants it sets web_host or SCENE_WEB_HOST.
+    assert resolve_web_host({}, {}) == "127.0.0.1"
 
 
 @pytest.mark.parametrize(
@@ -61,11 +64,20 @@ def test_explicit_blank_values_do_not_fall_back_to_all_interfaces():
 
 
 def test_launchers_preserve_blank_values_for_runtime_rejection():
+    # `${VAR-default}` and not `${VAR:-default}`: an operator who exports an
+    # empty SCENE_WEB_HOST has made a mistake, and it should reach
+    # resolve_web_host to be rejected rather than be replaced silently.
     root = Path(__file__).resolve().parents[1]
-    docker_launcher = (root / "scripts" / "start.sh").read_text(encoding="utf-8")
-    native_launcher = (root / "scripts" / "start_native.sh").read_text(
-        encoding="utf-8"
-    )
-    expected = "SCENE_WEB_HOST=\"${SCENE_WEB_HOST-0.0.0.0}\""
-    assert expected in docker_launcher
-    assert expected in native_launcher
+    for name in ("start.sh", "start_native.sh"):
+        launcher = (root / "scripts" / name).read_text(encoding="utf-8")
+        assert 'SCENE_WEB_HOST="${SCENE_WEB_HOST-' in launcher, name
+        assert 'SCENE_WEB_HOST="${SCENE_WEB_HOST:-' not in launcher, name
+
+
+def test_launchers_default_the_ui_to_loopback():
+    # The library default is loopback; a launcher that passed 0.0.0.0 would
+    # override it and put the unauthenticated UI back on every interface.
+    root = Path(__file__).resolve().parents[1]
+    for name in ("start.sh", "start_native.sh"):
+        launcher = (root / "scripts" / name).read_text(encoding="utf-8")
+        assert 'SCENE_WEB_HOST="${SCENE_WEB_HOST-127.0.0.1}"' in launcher, name


enkerewpo:
need to be tested

PR #246, opened 2026-09-09
fix(testing): run the post-restart pass once, with the real map gate

enkerewpo:
## What is wrong

PR #244 was meant to gate the post-restart scenario pass on a `/map` that actually carries known cells. It landed as a **second copy of the entire post-restart sequence** — `Drive a fixed mapping loop` → `Verify scene map persistence` → `Validate boot/shutdown lifecycle` → `Re-run scenario suite` — instead of an edit to the existing one. dev has been running that sequence twice ever since.

Two consequences, both visible in the last dev run (34382745214):

**The report has been showing the wrong numbers.** The duplicated re-run step writes `testing/logs/summary.json` and `$RUNNER_TEMP/scenarios.out`. Those belong to the *first* scenario pass. The surviving copy writes `summary.restart.json` / `scenarios.restart.out`, which is what the artifact upload and the report generator read. So the fresh-boot column of the published report was being overwritten by post-restart data.

**The gate was in the copy that did not matter.** The surviving step still waited on `ros2 topic info -v /map` reporting `Publisher count: [1-9]`. A publisher exists from the moment the node starts, so that wait returns immediately and proves nothing about the grid. The known-cells gate only ever ran in the duplicate.

## The failure on dev

The gated pass did its job: `/map: 156x185 grid, 6758 known cells`, then `RESULT: PASS — 16/16`. The run then went through a *second* mapping loop and a *second* lifecycle restart — a third boot of the stack — and failed six minutes into `waiting for a real non-robot scene object after lifecycle restart`, with `/api/state` holding nothing but `scene.object.robot_001`. That whole sequence exists only because of the duplication.

## The change

Delete the duplicated block and move the known-cells gate into the one that remains, keeping its `summary.restart.json` outputs. Net effect against the state before #244 is the gate and nothing else.

Also: both failure tails in that step could not open their log files (`tail: cannot open 'examples/webots/rbnx-boot/logs/scene.log'`). The step sets `working-directory: examples/webots`, so a path beginning with `examples/webots` resolved twice and every failure lost its diagnostics. Both are now anchored at `$GITHUB_WORKSPACE`.

## Validation

`.github/workflows/testing.yml` parses; the diff against the pre-#244 base is the gate plus the two path fixes, nothing else. The Webots suite on this PR is the real check — the same restart pass reported 16/16 on dev once it was gated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,17 @@ jobs:
           system/scene/tests/test_perception_vlm_geometry.py
           system/scene/tests/test_image_relations.py
 
+      # The goal planner and the message-shape rule are pure and import neither
+      # robonix_api nor the generated MCP types, so they run under the same
+      # lightweight dependency set. Without them here nothing on a pull request
+      # exercises the guards against a malformed occupancy grid, which is how
+      # those guards came to be missing in the first place.
+      - name: Test Scene goal planning and upstream message validation
+        run: >-
+          python3 -m pytest -q
+          system/scene/tests/test_goal_planner.py
+          system/scene/tests/test_malformed_grid_inputs.py
+
   check:
     name: Rust (fmt + clippy + build + tests)
     runs-on: ubuntu-22.04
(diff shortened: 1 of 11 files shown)


enkerewpo:
Landed directly on dev as 77325025 — a CI-only fix does not need a PR round trip. Closing.

PR #247, opened 2026-09-10
feat(rbnx): record which package commits a deployment actually ran

enkerewpo:
## Why

A deployment manifest pins packages by branch:

```yaml
service:
  - name: mapping
    url: https://github.com/syswonder/service-map-rbnx
    branch: feat/map-localizer
```

`install_from_github` does record a commit — but boot never goes through it. `rbnx boot` clones url-remote packages into `rbnx-boot/cache/` on first run and then reuses that checkout untouched forever. Nothing records what the cache holds, so the same manifest is not the same code twice and nobody can tell.

This is not hypothetical. A running deployment turned out to be **three commits behind its branch, with uncommitted edits applied by hand on top** — code that exists in exactly one place on one machine. Every measurement taken against that deployment had been attributed to the branch. Noticing took running `git status` inside the cache by hand; nothing about the run would ever have said so.

## What this does

Before spawning anything, boot writes `<manifest-dir>/rbnx-boot/deployment.lock`. The shape is taken from the lockfiles that had to solve this already:

- **`flake.lock`** keeps `original` and `locked` apart — what the input asked for versus what it resolved to. Here that is branch versus commit, and conflating them is the bug this file exists to expose, so they are separate blocks.
- **`Cargo.lock` / `package-lock.json`** carry a format version. So does this, and a reader never has to infer the shape.
- **All three are deterministic** — sorted entries, no timestamps, and the file is rewritten only when its content changes. A lock that rewrites itself on every boot has no readable diff.

```yaml
# Generated by `rbnx boot`. Do not edit.
version: 1
packages:
  - name: mapping
    original:
      url: https://github.com/syswonder/service-map-rbnx
      branch: feat/map-localizer
    locked:
      commit: df2ba718c0de1a5f...
      dirty: true
      dirty_digest: sha256:9c5e5cd...
```

A dirty checkout carries a **digest** of its uncommitted changes (`status` + `diff HEAD`, sha256), not just a flag: a boolean cannot tell two different local edits apart, and comparing runs across machines is the reason to record it at all. This is the same device the project's own reproducibility record uses for its overlays.

`path:` entries that are not their own git checkout record `commit: null` — they are versioned by whatever contains them.

Boot also warns for every dirty checkout, because a branch name cannot describe one.

## Deliberately written, not enforced

Boot still runs whatever the cache holds. Refusing to start on a moved branch needs a fetch on every boot and a policy for what to do about it — offline sites, pinned deployments, a robot in the field. Leaving evidence needs neither, and it is the half that has to exist before the other half can be argued about. A `--locked` mode is the obvious next step and is not in this change.

## Validation

- `cargo test -p robonix-cli` — 26 + 31 passed, one new case: it builds a real git checkout, writes the lock, asserts the recorded commit matches `git rev-parse HEAD` and that a clean tree is not flagged; then edits a file and asserts `dirty: true`.
- `cargo fmt --all` and `cargo clippy -p robonix-cli --all-targets -- -D warnings` clean.
- `tools/rbnx/README.md` documents the artefact under Subcommands.

### DIFF
--- a/tools/rbnx/README.md
+++ b/tools/rbnx/README.md
@@ -44,6 +44,20 @@ cargo install --force --path tools/rbnx --bin rbnx
 
 Run `rbnx <cmd> --help` for full flags.
 
+### What a boot leaves behind
+
+`rbnx boot` writes `<manifest-dir>/rbnx-boot/deployment.lock` before it spawns
+anything: one entry per package with the URL, the branch the manifest asked
+for, the commit the checkout was actually on, and `dirty: true` when that
+checkout had uncommitted changes.
+
+A manifest pins packages by branch, and a cache is cloned once and then reused
+untouched, so "the same manifest" is not the same code twice — a deployment was
+found running three commits behind its branch with hand-applied edits on top,
+and nothing about the run recorded it. The lockfile is that record. It is
+written, not enforced: boot still runs whatever the cache holds, and a dirty
+checkout is warned about rather than refused.
+
 ## Built-in system components
 
 `rbnx boot` launches each declared built-in component in dependency order:
--- a/tools/rbnx/README.md
+++ b/tools/rbnx/README.md
@@ -47,9 +47,27 @@ Run `rbnx <cmd> --help` for full flags.
 ### What a boot leaves behind
 
 `rbnx boot` writes `<manifest-dir>/rbnx-boot/deployment.lock` before it spawns
-anything: one entry per package with the URL, the branch the manifest asked
-for, the commit the checkout was actually on, and `dirty: true` when that
-checkout had uncommitted changes.
+anything. Each package records what the manifest asked for and what that
+resolved to, kept apart the way `flake.lock` keeps them:
+
+```yaml
+version: 1
+packages:
+  - name: mapping
+    original:
+      url: https://github.com/syswonder/service-map-rbnx
+      branch: feat/map-localizer
+    locked:
+      commit: df2ba718c0de1a5f...
+      dirty: true
+      dirty_digest: sha256:9c5e5cd...
+```
+
+Entries are sorted and nothing is timestamped, so the file changes when the
+deployment changes rather than every time it runs — a lock that rewrites itself
+on every boot is one nobody reads a diff of. A dirty checkout gets a digest of
+its uncommitted changes, not just a flag: a flag cannot tell two different
+local edits apart.
 
 A manifest pins packages by branch, and a cache is cloned once and then reused
 untouched, so "the same manifest" is not the same code twice — a deployment was
--- a/tools/rbnx/src/cmd/deploy.rs
+++ b/tools/rbnx/src/cmd/deploy.rs
@@ -184,6 +184,68 @@ fn resolve_entry_path(
 mod tests {
     use super::*;
 
+    fn git(dir: &Path, args: &[&str]) {
+        let ok = std::process::Command::new("git")
+            .args(args)
+            .current_dir(dir)
+            .output()
+            .expect("git")
+            .status
+            .success();
+        assert!(ok, "git {args:?} failed in {}", dir.display());
+    }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


enkerewpo:
## Test report

Built at `90abe6b`, run on macOS aarch64 with a fixture deployment, and cross-checked against the real package cache on the Ranger robot.

### Unit test

```
cargo test -p robonix-cli
26 passed; 0 failed; 1 ignored
32 passed; 0 failed; 1 ignored
```

including `cmd::deploy::tests::the_lockfile_records_the_commit_and_whether_the_tree_was_edited`.

### End-to-end

Two fixture packages as real git repos, pinned by branch in a deployment manifest, driven through `rbnx boot` three times. Boot exits non-zero afterwards (the fixture has no Atlas); the lockfile is written before anything spawns, so the artifact is produced either way.

| Check | Result |
|---|---|
| Lockfile written before spawn | pass |
| `version: 1` present | pass |
| `original` and `locked` kept apart | pass |
| Commit recorded per package | pass |
| Byte-identical across two unchanged runs | pass |
| No timestamp in the file | pass |
| Editing one cached checkout moves only that package's digest | pass |

The last row is the one that matters. After editing `beta`, its digest moved from `f35282b7…` to `97948e9f…` while `alpha`'s stayed put:

```yaml
  - name: beta
    original:
      url: file:///tmp/lockfile-test/remotes/beta
      branch: master
    locked:
      commit: 4a2ff95db9ae6fb193ce05c1200a0ce4fbd5b974
      dirty: true
      dirty_digest: sha256:97948e9f4f1642aed3755c54f656e4d975384528941f87d23666522fc9e90f58
```

### A finding, and a fix for it in this branch

On the Ranger deployment, 7 of 12 cached packages reported dirty, and in every case the whole change was:

```
$ git -C rbnx-boot/cache/primitive-livox-mid360-lidar-rbnx status --porcelain
?? logs/
```

The file inside is `logs/codegen.log`, written by `robonix-codegen`. Scribe resolves its log directory as `$SCRIBE_LOG_DIR` or `./logs` (`system/scribe/src/lib.rs:281`), rbnx runs codegen with the package directory as the working directory, and nothing sets that variable — so the log lands in the checkout root, beside the sources, instead of under `rbnx-build/` with every other build artefact. Packages gitignore `rbnx-build/` but not `logs/`, so `dirty: true` fires on the first build and never clears, and the digest then tracks log filenames rather than code.

That is a log-placement bug, not a lockfile one: filtering directory names inside `git_state()` would be papering over it, and rbnx would be second-guessing what belongs to the operator. The fix is to put the log where the convention already says it goes. `build_codegen_cmd` now takes a log directory and sets `SCRIBE_LOG_DIR`; the four call sites point it at `rbnx-build/logs`, the same path `run_package.rs` already uses.

After the fix, a freshly built checkout carries only `?? rbnx-build/`, which packages already gitignore, and `dirty: true` again means what it says.

### Merge conflict

`dev` gained a `README.md` section for `llm_callable` in the same place this branch added the lockfile section. Both are kept: the `rbnx tools` visibility note first, the boot artefact section after it. No code conflict.

### Status

CI was green on `90abe6b` (12/12). The conflict resolution and the codegen log-directory fix are prepared and tested locally; pushing them is pending.

enkerewpo:
## Test for the codegen log fix

Same package, same command, only the two patched files differing.

| | `logs/` in the checkout root | `rbnx-build/logs/` |
|---|---|---|
| before (`90abe6b`) | **present**, holding `codegen.log` | absent |
| after (`73182547`) | absent | **present**, holding `codegen.log` |

```
$ cd services/memsearch && rm -rf rbnx-build logs && rbnx codegen
```

Reverting only `tools/rbnx/src/cmd/{codegen,docs}.rs` to `90abe6b` reproduces the old placement, so the change is doing the work rather than something else in the merge.

The `grpcio-tools` error the command ends on is unrelated (a missing Python dependency on the test machine); codegen gets far enough to write its log either way, which is the part under test.

`cargo test -p robonix-cli` stays at 58 passing after the change.

PR #248, opened 2026-09-10
feat(scene): draw the semantic map with rerun and give the UI a sidebar

enkerewpo:
## Problem

Scene drew its own map. A three.js page pulled the library from a CDN and needed WebGL, and a canvas floor plan sat beside it. On a headless robot neither is available, so the 3D view rendered nothing there, and keeping both pages working was viewer maintenance that has nothing to do with a scene graph. The pages also had no links between them: reaching the annotation view from the map meant editing the address bar, and nothing on any page said the other views existed.

Scene's semantic map in the rerun viewer

*A live Webots office session. Each object is its own point cloud carrying one label; the occupancy grid is drawn underneath as the floor, white where the robot has observed it and dark at the walls; the orange polygon is the robot's own footprint with its heading. Relation edges are drawn thin and pale on purpose — they are context, and the clouds are the evidence. The box each object reports is logged but hidden, and switched on from the entity tree when a detection needs checking.*

## Change

The map pages are rerun views. Scene logs what it knows and rerun draws it. What is logged is the annotated map, not raw sensor data: one point cloud per object coloured by instance and carrying its class label, a second cloud in the colours the camera saw, the occupancy grid as a textured floor with one colour per cell meaning, one edge per relation, and the robot's own footprint polygon with a heading arrow.

The 3D page is the landing page, because whether the perception is any good is the question this UI is opened to answer. The 2D page is the same annotations seen from above, drawn in the grid's own pixel coordinates so they line up with the map rather than with a second drawing of it. Read that way, an object either sits in the room it belongs to or it does not, and a relation either crosses a wall or it does not. Every page now shares a sidebar.

Each object's reported box is logged but hidden by default. It is the perception layer's own extent estimate, it is often wrong, and a wireframe around every object buries the points that are the actual evidence. It is switched on from the entity tree when a detection needs checking.

Two recordings on two gRPC ports (9876 and 9877) are served by one web viewer (9090), because a blueprint belongs to an application and the viewer takes its data source from the query string.

## Why rerun

[rerun](https://rerun.io/) ([rerun-io/rerun](https://github.com/rerun-io/rerun), Apache-2.0) is an open-source SDK and viewer for multimodal robotics data. It supplies four things this PR would otherwise have had to build and keep working:

**The viewer, and the server that serves it.** The page being replaced pulled three.js from a CDN and needed WebGL in the browser. A headless robot has neither, so that page rendered nothing exactly where a viewer is most useful. rerun ships the viewer as a WASM bundle, an HTTP server for it, and the wire protocol between them, so Scene logs entities and links to the viewer instead of maintaining one.

**A data model that matches what Scene has.** rerun is entity-and-component, not draw calls. Each object becomes three entity paths — `rgb_pcd/<id>` for the colours the camera saw, `sem_pcd/<id>` for one colour per instance, `bbox/<id>` for the extent estimate — and the viewer's entity tree makes each one independently visible. That is why the debug boxes can be a layer the reader switches on rather than a UI feature we implement: the blueprint hides `/map/objects/bbox` by default and the tree switches it back on.

**A time axis for free.** `set_time` puts every tick at a point on a timeline, so dragging it replays the map filling in as the robot drove. That is what separates a detection that persisted from one that appeared for a single frame, and it is not something a hand-rolled viewer gets without real work.

**A declarative layout.** The blueprint is sent with the data, so a page opens on the right view with the panels collapsed, with no viewer-side configuration for an operator to get wrong.

### It is the same shape of data the official examples show

rerun's own [examples](https://rerun.io/examples) include several that are structurally what Scene produces, which is a good sign that the model fits rather than that we are bending it:

- [ARKitScenes](https://rerun.io/examples/spatial-computing/arkit_scenes) — posed RGB-D frames back-projected into a room, with labelled 3D boxes over it. This is the closest to Scene's semantic map.
- [RGBD](https://rerun.io/examples/robotics/rgbd) and [live depth sensor](https://rerun.io/examples/robotics/live_depth_sensor) — the depth-to-point-cloud path.
- [ROS 2 bridge](https://rerun.io/examples/integrations/ros2_bridge) — for reference only. Scene deliberately does not take this path: the sink reads the registry, not ROS topics, so the viewer shows the annotated map rather than raw sensor data.

The three-layer per-object layout follows [DualMap](https://github.com/Eku127/DualMap), which uses rerun the same way and is already a selectable Scene backend.

### What it costs, stated honestly

- The viewer's own chrome (top tab bar, timeline) is dark and stays dark: rerun 0.37's web backend ignores `SetTheme`, so the light theme applies to the view contents and Scene's own shell, not to that strip.
- rerun-sdk declares `numpy>=2`, which the perception stack cannot take. It is therefore an **optional extra**, not a dependency, and the x86 image installs it with `--no-deps` and pinned imports. Native installs keep the built-in pages and nobody porting Scene has to solve for it.

## Native deployments are unchanged

`web_viewer` in the Scene config, or `SCENE_WEB_VIEWER`, selects `auto`, `rerun` or `builtin`. `auto` uses rerun when it is importable, which is true in the x86 docker image and false natively, so a native install keeps exactly the pages it had. rerun is an optional extra rather than a dependency: a port to a new board has to be able to bring Scene up without solving for it, and rerun-sdk declares `numpy>=2`, which the perception stack cannot take. The x86 image installs it with `--no-deps` and pinned imports. Setting `rerun` where it cannot start fails the start with the reason rather than serving an empty frame. The built-in pages stay reachable at `?bare=1` either way.

## Boundary

The sink only reads. It takes a registry snapshot once a second and publishes it. Object persistence, annotations and the map lifecycle are untouched, and a viewer that cannot start does not stop the map service.

## Validation

- Full stack on the Webots ConceptGraphs profile (`robonix_manifest.scene-eval.cg.nav.yaml`), driven by the standard Explore skill. Verified in the live recording that every layer is logged: `/map/floor`, `/map/objects/{rgb_pcd,sem_pcd,bbox}/<id>`, `/map/relations`, `/map/robot/{footprint,heading}`, and the `/map2d` equivalents.
- Every route checked through the running service: `/`, `/2d`, `/cam`, `/user`, `/3d` and each `?bare=1` form.
- `pytest system/scene/tests/{test_rerun_sink,test_web_live_views,test_scene_graph,test_vlm_inference_cache,test_perception_vlm_geometry,test_image_relations,test_web_binding}.py`: 95 passed.
- `test_rerun_sink.py` is added to the CI Scene job.

No Rust changes. No capability contract, generated code, or migration changes.

## Also

`testing/export_scene_pointcloud.py` writes the same objects to a coloured PLY and an `.rrd` for offline comparison against ground truth.


enkerewpo:
## Test report

Run on a live Webots office session (TIAGo, `robonix_manifest.scene-ui-local.yaml`, concept_graphs backend, rerun 0.37.1), driven on a manual tour.

### Automated

`pytest system/scene/tests` — **401 passed, 9 skipped, 0 failed**, after merging `dev` (one conflict: `dev` inlined the web-port fallback chain this branch had already extracted into `_web_port_from`; kept the function, since the viewer now reads the port before the server is built and a duplicated chain can disagree with itself about whether the UI exists).

### Pages and endpoints

All 200: `/` `/2d` `/3d` `/regions` `/maps` `/cam` `/logs` `/rerun`, and `/api/state` `/api/objects3d` `/api/viewer` `/api/camera` `/api/maps` `/api/regions` `/api/logs`.

### Operations, end to end against a live session

| operation | result |
|---|---|
| rename object | `{"ok":true,"label":"测试椅"}`, reflected in `/api/state`; non-ASCII labels survive |
| delete object | `{"ok":true}`, gone from `/api/state` and **did not come back** on later ticks — the detector hook fired, not just the registry |
| focus view | `{"ok":true}` |
| object views (photos) | 200 |
| region create / list / delete | created `测试区` (`anno.0ccc0e3f`), listed, deleted |
| map save | `saved spatial map ui_test_map; rtabmap backup completed; sqlite_backup ok`, 21 objects persisted |
| map list | shows the saved map |
| flush | `{"ok":true,"deleted":40,"persisted":true}`; perception repopulated afterwards, as intended |

Two rejections were the API being right, not wrong: `flush` refused a stale request with `Scene map epoch changed: requested ('', None), current ('ui_test_map', -1)` after a map save, and `regions` rejected `kind: room` naming the two it accepts.

### Fixed in this PR

**The landing page could never show the viewer.** It chose its layout from `rerun_sink.ready`, but bring-up had moved off the boot path: `ready` only turns true once `/rerun` is opened, and `/rerun` is only fetched by the frame the page was declining to render. A deployment with rerun installed served the built-in canvas for ever, and silently — that fallback is also the one a native install legitimately takes. Split into two questions: `available` (does this deployment have a viewer; `find_spec`, no ports bound) decides layout, `ready` (is one running) stays with the routes that forward to a live server.

**The map drew into a corner** — 640x360 inside a 969x902 frame. Two inline styles, each beating the stylesheet. The viewer writes `position: relative` onto the host as it starts, which left `inset: 0` inert and collapsed the host onto its content, and that content was a canvas sized to a default precisely because the host had no height to measure. The host is now sized in percent, which survives the viewer's `position`; the canvas itself is pinned with `!important` over the `width: 640px` `@rerun-io/web-viewer` writes onto it. Measured after: host 969x902, canvas 969x902, backing store 1937x1804 at dpr 2.

Both are pinned by tests: `test_viewer_gate.py`, `test_rerun_host_css.py`.

### A note for anyone debugging the viewer

A long investigation in `NOTES-open-issues.md` chased a "blank" rerun canvas. Most of it was measuring the harness. An automated browser tab is hidden, `requestAnimationFrame` never runs, eframe never paints, the canvas stays at its 300x150 default — and rerun opens its gRPC stream from inside that same update loop, so no request is made either, which reads exactly like data never arriving. Forcing one frame took the same canvas to 3308x1882. The three signals that investigation trusted are all invalid in a hidden tab; measure the DOM instead. The GIL deadlock in `serve_grpc` is real and is still why bring-up is lazy.

### Not covered

Interactive rerun behaviour — entity-tree toggles, timeline scrubbing, click-to-select crossing the frame boundary — needs a visible browser and a person. Everything above is automated or API-level.

enkerewpo:
## CI is green, and what it took

Two checks were failing. Neither was introduced by the changes above, but both are in this PR.

### Docs check (English-only) — red since 19 September

654 lines of CJK, in three groups.

**495 of them were working notes.** Five `NOTES-*.md` files recording what was tried on particular evenings. That is scratch, and a repository whose code and docs are English is not where it goes. Removed from the history, kept on disk, ignored.

**151 were the interface copy**, a `{"en": ..., "zh": ...}` table inside `web.py`. The table is not the problem — putting it in a Python module is. Translations are content: no logic, edited by whoever speaks the language, and a translator should not have to read Python to change a word. They are `web_assets/strings_en.json` and `strings_zh.json` now, one file per language, composed at import, English as the reference with per-key fallback.

Keeping each file to one language made the language switch awkward, since its label is the language you are *not* in — the English table used to carry the Chinese word for Chinese. The switch renders from the other table through a new `data-i18n-other`, so `strings_zh.json` is Chinese and nothing else.

**The last 8** were comments using a Chinese room name as their example, and two tests that restated interface copy in order to assert on it. The comments make the same point in English. The tests read the copy files, so the wording lives in one place. The two non-ASCII search fixtures are written as escapes: the test is that a query whose text is non-ASCII resolves by class, and it stays that test while the source around it stays English.

### Generic runtime portability audit

`_navigator` stamped `goal.header.frame_id = "map"` on every navigation goal. That is what this deployment calls its world frame, not what every deployment calls it, and a goal sent in a frame the robot does not use is a robot driving to the wrong place — silently, because the pose numbers are all plausible.

Scene already knows the answer: `_SelfTracker.world_frame_id`, resolved from the first transform and handed to the detectors as `world_frame_fn`. `mcp_tools` receives it through `attach_state` now, the way it receives the registry and the hub, as a callable rather than a value — the frame is empty until the first transform arrives, and a value captured at attach time would be that emptiness for ever. With no frame resolved the goal is refused rather than guessed.

### After

All checks pass. `pytest system/scene/tests` — **401 passed, 9 skipped**. The UI was re-checked live afterwards: the language switch reads from the other table correctly (shows 中文 while `html lang="en"`), and the viewer still sizes right — host 969x902, canvas 969x902, backing store 1937x1804 at dpr 2.

### Still open, and not for this PR

The object list in a live run shows one continuous run of office shelving as `cabinet_006/010/012/013`, `shelf_005/014`, `picture_frame_001..005`. The merge criterion is point-cloud overlap, and two adjacent segments of one long object do not overlap each other, so no distance threshold reaches them — `max_merge_dist_m` never gets a say. The `association` block the eval manifests carry (`same_class_centroid_max_m`, `exact_duplicate_*`, ~20 knobs written for exactly this) is read by no code on this branch, on #243, or on `dev`; its implementation is on `scene-object-management-177` / `scene-benchmark-harness`, neither merged. Worth a decision separately from this PR.

PR #251, opened 2026-09-15
feat(executor): support overlapping result verification

KouweiLee:
## Summary

This PR adds optional overlap between capability execution and result verification.

When overlap is enabled, Executor no longer blocks the remaining RTDL tree while a successful capability result is being verified. Instead, it emits a `VERIFYING` state, continues eligible sibling execution, and finalizes the affected nodes after verification completes.

## Changes

- Add an executor-wide `verification.overlap` option, defaulting to `false`.
- Add the `VERIFYING` RTDL node state.
- Run matched verification rules in the background when overlap is enabled.
- Allow sequence and parallel siblings to continue while verification is pending.
- Propagate verification failures to the affected leaf and its parent operators without changing already-executing siblings.
- Wait for all pending verification work before emitting `plan_complete`.
- Include verification failures in the plan's `any_failed` result.
- Preserve the final verification outcome of a leaf that entered `VERIFYING`, even if the plan is subsequently canceled.
- Update Executor, Pilot, verifier documentation, and the Webots example configuration.

## Configuration

```yaml
system:
  executor:
    verification:
      overlap: true
      rules:
        - target_contract_id: robonix/service/navigation/navigate
          target_provider_id: simple_nav
          verifier_provider_id: scene_verifier
          verifier_args:
            scene_provider_id: scene
```

## Execution behavior

With overlap: true:
1. A successful matched capability call emits VERIFYING with its result.
2. Verification continues in the background.
3. Eligible sequence or parallel siblings can continue executing.
4. The leaf later emits exactly one terminal state: SUCCEEDED or FAILED.
5. Parent operators remain VERIFYING until their descendant verification work finishes.
6. plan_complete is emitted only after all final verification outcomes are available.

When overlap is omitted or set to false, verification remains synchronous.


1mujue (commented):
Everything is fine EXCEPT for service.rs, since there are too many new codes and changes. Once it was tested both locally and on GitHub and then succeeded, I think it won't cause any problem.

1mujue (approved):
It is ok to merge.

